PR #911, opened 2025-04-23
RISDEV-7552 Add nuxt-security for CSP, other security headers

MasterCarl:
This PR adds the nuxt-security module, which adds content-security-policy, permissions-policy, x-frame-options, and referrer-policy headers


hmones (approved):
Looks good to me

PR #916, opened 2025-04-28
RISDEV-7182 Replace unsorted with case-specific sort

MasterCarl:
This PR replaces the default sort with a case-specific one, descending by date. It adds an integration test case to assert the sort order for mixed results.


backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/PaginationParamsConverter.java
```diff
@@ -63,4 +65,12 @@ private static boolean isSupportedField(
...
     };
   }
+
+  private static Sort getDefaultSort(MappingDefinitions.ResolutionMode resolutionMode) {
+    return switch (resolutionMode) {
+      case ALL -> Sort.by(Sort.Direction.DESC, "DATUM");
+      case NORMS -> Sort.by(Sort.Direction.DESC, Norm.Fields.NORMS_DATE);
+      case CASE_LAW -> Sort.by(Sort.Direction.DESC, CaseLawDocumentationUnit.Fields.DECISION_DATE);
+    };
+  }
```
MagdaN, line 75: Do we have more code that distinguishes between document kinds? If yes then we should maybe consider to have a special configuration classes for this?

backend/src/test/java/de/bund/digitalservice/ris/search/integration/controller/api/values/SortingTestArguments.java
```diff
@@ -20,7 +24,21 @@ public static Stream<Arguments> provideSortingTestArguments() {
...
+        result -> {
+          JSONObject jsonObject = new JSONObject(result.getResponse().getContentAsString());
+          ArrayList<String> dates = new ArrayList<>();
+          for (Object member : jsonObject.getJSONArray("member")) {
+            var item = ((JSONObject) member).getJSONObject("item");
+            if (item.has("decisionDate")) {
+              dates.add(item.getString("decisionDate"));
+            } else if (item.has("legislationDate")) {
+              dates.add(item.getString("legislationDate"));
+            }
```
MagdaN, line 37: Same here

MagdaN (approved):
Looks good, but maybe we should consider moving logic that returns different values based on document kind to a config class.

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/PaginationParamsConverter.java
MasterCarl: Good call, I've moved the code to a renamed class `SortingDefinitions`, placed next to `MappingDefinitions`

PR #919, opened 2025-04-30
Support highlighting file numbers

MasterCarl:
This PR adds support for highlighting of file numbers in case law search results.

In order to simplify the logic, the textMatch key conversion (from snake_case to camelCase, dropping `.text` suffix where used) is now done in all cases, not just in the AllDocumentsSearchController.

Test cases have been added. In the edge case that a file number is input without any whitespace or special characters, the highlighting is not expected.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/utils/RisHighlightBuilder.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/utils/RisHighlightBuilder.java
@@ -66,6 +66,12 @@ public final class RisHighlightBuilder {
   private static HighlightBuilder addCaseLawFields(HighlightBuilder builder) {
     CASE_LAW_HIGHLIGHT_CONTENT_FIELDS.forEach(builder::field);
     builder.field(getFieldBasic(CaseLawDocumentationUnit.Fields.ECLI));
+    /*
+     * Note that this uses the .text variant of file_numbers
+     * */
+    builder.field(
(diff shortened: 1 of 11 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/utils/RisHighlightBuilder.java
```diff
@@ -66,6 +66,12 @@ private static HighlightBuilder addNormsFields(HighlightBuilder builder) {
   private static HighlightBuilder addCaseLawFields(HighlightBuilder builder) {
     CASE_LAW_HIGHLIGHT_CONTENT_FIELDS.forEach(builder::field);
     builder.field(getFieldBasic(CaseLawDocumentationUnit.Fields.ECLI));
+    /*
+     * Note that this uses the .text variant of file_numbers
+     * */
+    builder.field(
+        getFieldBasic(CaseLawDocumentationUnit.Fields.FILE_NUMBERS_TEXT)
+            .numOfFragments(5) /* Support highlighting multiple (partial) file numbers */);
```
MasterCarl, line 74: @hmones I've added a special case for highlighting file numbers since our pair programming session, just like we already did for ECLIs. The reason was that setting the default `numOfFragments` to something higher than 1 meant that multiple matches for e.g. headline would start to appear.

hmones: Well done, I think that approach is good.

hmones (approved):
Looks great :tada:

PR #922, opened 2025-04-30
Use correct Docker image tag

MasterCarl:
Sometimes, E2E pipelines may not use the most recent frontend or backend image, if that image has been built in the previous step. This PR changes this to use the github.sha value, if applicable.
The frontend E2E pipeline will still use the PR-branch or main backend image, and vice versa. This should be acceptable for most cases.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -26,9 +26,103 @@ env:
   SENTRY_RELEASE: ${{ github.sha }}
 
 jobs:
+  check-frontend:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Node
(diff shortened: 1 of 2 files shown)


hmones (changes requested):
@MasterCarl The idea behind the switch earlier from the github sha to the branch name was that the storage for the github images as going really high and hard to maintain so many images that needs to be deleted and are there all the time. That's why I switched to persisting the image with the branch name and consulted Hanna as well on that back then. Maybe we can debug the issue you were facing, because it normally shouldn't happen an older image is pulled since each push would override the older image before the e2e tests run. We can debug it together if you like

MasterCarl:
@hmones Okay, let's take a look. I'm not creating any new Docker image tags though, since every build pipeline pushes the branch tag as well as the SHA.

PR #924, opened 2025-05-05
Change ObjectStorage method naming, exceptions

MasterCarl:
This PR changes the naming of ObjectStorage methods to be more in line with standard object storage terms, like using "object" instead of "file" and "key" instead of "path".
It also changes the RetryableObjectStoreException to a more generic ObjectStoreServiceException, with a custom NoSuchKeyException to differentiate such cases.

The `get` method is removed to enforce more specific exception handling. Up to this point, any kind of exception thrown in this method would cause clients to get a `404` error code.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/exception/NoSuchKeyException.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/exception/NoSuchKeyException.java
@@ -1,6 +1,6 @@
 package de.bund.digitalservice.ris.search.exception;
 
-public class NoSuchKeyException extends ObjectStoreServiceException {
+public class NoSuchKeyException extends Exception {
   public NoSuchKeyException(String message, Throwable cause) {
     super(message, cause);
   }
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/exception/NoSuchKeyException.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/exception/NoSuchKeyException.java
@@ -1,0 +1,7 @@
+package de.bund.digitalservice.ris.search.exception;
+
+public class NoSuchKeyException extends ObjectStoreServiceException {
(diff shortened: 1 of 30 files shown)


andrew-cenkner-digitalservice (approved):

backend/src/main/java/de/bund/digitalservice/ris/search/exception/NoSuchKeyException.java
```diff
@@ -0,0 +1,7 @@
+package de.bund.digitalservice.ris.search.exception;
+
+public class NoSuchKeyException extends ObjectStoreServiceException {
+  public NoSuchKeyException(String message, Throwable cause) {
```
ekl176, line 4: I don't think we want to extend from the ObjectStoreServiceException here, since it is not an exception of service availability (retry-able)

backend/src/main/java/de/bund/digitalservice/ris/search/service/ImportService.java
```diff
@@ -59,7 +59,7 @@ public void lockAndImportChangelogs(
           importChangelogs(indexService, changelogBucket, lastSuccess, lastSuccessFilename);
         }
         indexStatusService.unlockIndex(lockfile);
-      } catch (RetryableObjectStoreException e) {
+      } catch (ObjectStoreServiceException e) {
         logger.error("Import process failed due to a retryable error. Removing lock", e);
```
ekl176, line 63: this shouldn't catch the NoSuchKeyException

ekl176 (changes requested):

backend/src/main/java/de/bund/digitalservice/ris/search/service/ImportService.java
MasterCarl: Good catch (no pun intended)!

PR #925, opened 2025-05-05
Run main Spring application for data ingestion

MasterCarl:


backend/src/main/java/de/bund/digitalservice/ris/search/importer/ImportTaskProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+    } catch (IllegalArgumentException ex) {
+      logger.error(ex.getMessage(), ex);
+      return ERROR_RETURN_CODE;
+    }
+  }
+
+  @NotNull
+  public static List<String> parseTargets(String[] args) {
+    List<String> targets = new ArrayList<>();
+    IntStream.range(OK_RETURN_CODE, args.length)
```
andrew-cenkner-digitalservice, line 58: I think the use of OK_RETURN_CODE was an auto refactor that doesn't make sense. It should start at 0.

andrew-cenkner-digitalservice: This IntStream loop feels a lot more clunky than
```
for(int i=0;i<args.length-1;i++){
  if(args[i].equals(TASK_ARGUMENT)){
    String target = args[i + 1];
    targets.add(target);
  }
}
```
is there a reason to go with the non standard IntStream forEachOrdered?

backend/src/main/java/de/bund/digitalservice/ris/search/importer/ImportTaskProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+        .filter(i -> args[i].equals(TASK_ARGUMENT))
+        .forEachOrdered(
+            i -> {
+              try {
+                String target = args[i + ERROR_RETURN_CODE];
```
andrew-cenkner-digitalservice, line 63: Using ERROR_RETURN_CODE looks like another auto refactor mistake.

backend/src/main/java/de/bund/digitalservice/ris/search/importer/ImportTaskProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+  public int runTask(String target) {
+    return switch (target) {
+      case "import_norms" -> {
+        try {
+          normIndexSyncJob.runJob();
+        } catch (ObjectStoreServiceException e) {
+          logger.error(e.getMessage(), e);
+          yield ERROR_RETURN_CODE;
+        }
+        yield OK_RETURN_CODE;
```
andrew-cenkner-digitalservice, line 82: I prefer the yield to be directly under the `runJob();` for two reasons
* It's logically easier to understand not splitting the mental flow of one case across the other case
* Following this pattern (yield or return are either inside **both** try and catch or **neither**) you can catch some bugs (like forgetting a return case for one of them) at compile time

Still, I would put this up to a style choice so it's okay if you want to leave it as is.

doc/adr/0013-s3-opensearch-Synchronization.md
```diff
@@ -0,0 +1,91 @@
+# 13. S3 Buckets and OpenSearch Synchronization
+
+Date: 2025-05-05
+
+## Status
+
+In discussion
```
andrew-cenkner-digitalservice, line 7: Given the previous lean coffee also with Magda and Eric this can be switched to Decided.

andrew-cenkner-digitalservice (approved):

backend/src/main/java/de/bund/digitalservice/ris/search/importer/ImportTaskProcessor.java
MasterCarl: Ah yes, that makes complete sense. I've changed the structure to match your suggestion.

PR #929, opened 2025-05-06
list keys with incomplete prefix

ekl176:
walk the parent dir to list all keys that contain a prefix in filesystem implementation


MasterCarl (approved):
Looks good, thanks for adding this feature!

PR #930, opened 2025-05-07
RISDEV-7621 add different searches to browser history

MasterCarl:
This PR changes `router.replace` to `router.push` when propagating Pinia state changes to the browser History API, and adds logic to enable the store to update according to the URL.

In order to prevent cycles where the store updates the route and the route change triggers a store update, logic is added to ensure that the store ignores route changes triggered by it.

Test cases have been added, including for PostHog reporting.

### DIFF
--- a/frontend/src/stores/searchParams/index.ts
+++ b/frontend/src/stores/searchParams/index.ts
@@ -141,10 +141,12 @@ export const useSimpleSearchParamsStore = defineStore(
       () => route.query,
       async (newQuery) => {
         // check whether the store already has the values, in order to prevent loops
-        const updatedQueryWasSetByStore = _.isEqualWith(
-          toRaw(routerQuerySetByStore.value),
+        const updatedQueryWasSetByStore = _.isEqual(
+          _.mapValues(
+            toRaw(routerQuerySetByStore.value),
+            (value) => value?.toString(), // numbers in query are represented as strings
+          ),
           newQuery,
-          (a, b) => a.toString() == b, // numbers in query are represented as strings
         );
 
         if (!updatedQueryWasSetByStore) {
--- a/frontend/src/stores/searchParams/index.ts
+++ b/frontend/src/stores/searchParams/index.ts
@@ -8,7 +8,7 @@ import {
   getInitialState,
   omitDefaults,
 } from "./getInitialState";
-import { ref, watch, computed } from "vue";
+import { computed, ref, watch } from "vue";
 import { sortMode } from "@/components/types";
 import type { LocationQuery } from "#vue-router";
 import _ from "lodash";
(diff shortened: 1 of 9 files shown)


frontend/src/stores/searchParams/index.ts
```diff
@@ -92,23 +97,69 @@ export const useSimpleSearchParamsStore = defineStore(
...
       pageNumber.value = initialState.pageNumber;
       category.value = initialState.category;
       itemsPerPage.value = initialState.itemsPerPage;
       sort.value = initialState.sort;
       court.value = initialState.court;
-      dateParams.$reset();
+      dateParams.reset(initialState);
     }
 
     /*
```
hmones, line 115: If I were you I'd get rid of the extra comments, usually according to clean code principles, a code should have structure and names that are self explanatory and wouldn't need comments to illustrate what it does. That's the case in other files too, but I thought to leave the comment here anyhow for you to consider if you like.

frontend/src/stores/searchParams/index.ts
```diff
@@ -92,23 +97,69 @@ export const useSimpleSearchParamsStore = defineStore(
...
+          ),
+          newQuery,
+        );
+
+        if (!updatedQueryWasSetByStore) {
+          reinitializeFromQuery(newQuery);
+          console.log(newQuery);
+        }
+      },
+    );
```
hmones, line 162: To me, the piece of code took a little while to understand and it needed comments to explain.
I would simplify it and make small functions to explain what happens and make it easier to debug and give single responsibility to each method of the code. Here's my suggestion or something in the same line of thought:

```suggestion
function normalizeQuery(query: Record<string, any>): Record<string, string> {
  return Object.fromEntries(
    Object.entries(query).map(([k, v]) => [k, v?.toString() ?? ""])
  );
}

function isSameQuery(
  storeQuery: Record<string, any>,
  routeQuery: Record<string, any>
): boolean {
  return _.isEqual(
    normalizeQuery(toRaw(storeQuery)),
    normalizeQuery(routeQuery)
  );
}

watch(
  () => route.query,
  (newQuery) => {
    if (!isSameQuery(routerQuerySetByStore.value, newQuery)) {
      reinitializeFromQuery(newQuery);
      console.log("The query was set by the user and is equal to:", newQuery);
    }
  }
);
```

hmones (commented):
I left some comments, and github actions also left some comments to check :) Let's clear them out and we're good to go :)

frontend/src/stores/searchParams/index.ts
MasterCarl: Clean code doesn't forbid all comments ;) I would keep a comment detailing that the intention is to prevent infinite loops.
Your suggestion does make sense though!

frontend/src/stores/searchParams/index.ts
MasterCarl: Good call, I've refactored the code to make the intention clearer.

PR #931, opened 2025-05-07
Risdev 7681 make indexing interuptable

andrew-cenkner-digitalservice:
Refactor ImportService to have 2 instances instead of being a singleton. As a result IndexingState is much more clear without things that belong in the ImportService constructor.


backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexStatusService.java
```diff
@@ -23,31 +23,28 @@ public IndexStatusService(PortalBucket portalBucket) {
...
       return true;
     } else {
       return false;
     }
   }
 
-  public boolean canLock(IndexingState state) {
-    PersistedIndexingState persistedState = state.getPersistedIndexingState();
-    if (persistedState.lockTime() == null) {
+  public boolean canLock(String StatusFileName, IndexingState state) {
```
MasterCarl, line 36: This should probably be lowercase

MasterCarl (approved):
Looks good. I think this makes the importer logic much easier to understand.

PR #932, opened 2025-05-07
RISDEV-7560 Remove talisman and introduce github secrets check

hmones:
This PR replaces the use of **Talisman** with **GitHub Secrets Push Prevention** to simplify the development workflow and enhance security. The decision aligns with the repository's transition to being public, leveraging GitHub's built-in features for secret detection.

The following has changed as part of this PR
1. The pipeline for secrets check using talisman has been removed since it's prevented originally from being pushed.
2. The talisman configuration file has been removed.
3. The details of the decision and consequences have been documented in an ADR.
4. The lefthook has been simplified and updated accordingly.


doc/adr/0013-replace-talisman-with-github-secrets-check.md
```diff
@@ -0,0 +1,61 @@
...
+1. **Enable GitHub Secrets Push Prevention**:
+   - This feature is enabled in the repository settings and prevents secrets from being pushed to the repository.
+
+2. **Define Bypass Patterns**:
+   - Specific patterns to bypass the secrets check can be defined directly in the GitHub repository settings, eliminating the need for additional configuration files in the repository.
+
+3. **Override Permissions**:
+   - To handle cases where pushing without restrictions is necessary, we have:
+     - Made all members of the **Portal Team** maintainers.
+     - Restricted pushing without checks to **admin accounts only**.
```
MasterCarl, line 42: I don't think this change is necessary. A bypass doesn't appear to happen automatically, but needs to be selected from a list of reasons:

> By default, anyone with write access to the repository can choose to bypass push protection by specifying one of the bypass reasons outlined in the table.

https://docs.github.com/en/code-security/secret-scanning/introduction/about-push-protection#how-push-protection-works

If push protection erroneously flags e.g. a test value as a sensitive secret, an override should be possible using personal accounts, so that there is an audit trail.

hmones: We can maybe test it on this branch if you like. My perception was that it would push still because they have write permissions, but we can test together to see how it works in practice.

hmones: I think the MYSQL_CONNECTION_STRING Is a non-provider pattern according to this page https://docs.github.com/en/enterprise-cloud@latest/code-security/secret-scanning/introduction/supported-secret-scanning-patterns#non-provider-patterns. We may need to look at this probably and if there's something we can identify as a pattern ourselves then we can add them as new patters in the new patterns list, but what I understood is that they check if the secret is something that is integrated in your app or so and it uses copilot AI to check (we have that enabled). We could take that to the security guild also and check with them. What do you think? And meanwhile we keep the talisman active

frontend/.env.prototype
```diff
@@ -7,4 +7,7 @@ NUXT_PUBLIC_BACKEND_URL=http://localhost:8090
 # overrides runtimeConfig.ssrBackendUrl in nuxt.config.ts (used for SSR)
 NUXT_SSR_BACKEND_URL=http://localhost:8090
 # overrides runtimeConfig.public.authEnabled
-NUXT_PUBLIC_AUTH_ENABLED=false
\ No newline at end of file
+NUXT_PUBLIC_AUTH_ENABLED=false
+
+MYSQL_CONNECTION_STRING=Server=http://mysql.neuris.local;Port=8002;Database=neuris_demo;Uid=neur;Pwd=fk2df1sdxSdjsd;
+ADAFRUIT_IO_KEY=1a2b4517692b5ad9
```
hmones, line 13:
```suggestion
```

MasterCarl (approved):
Looks good, as discussed in the team

PR #936, opened 2025-05-12
Replace Axios search by useFetch

MasterCarl:
This PR replaces the Axios searchService through useAsyncData with a more Nuxt-standard `useFetch` implementation.
The advantage is that useFetch handles passing client credentials to the server in SSR mode. It is also more concise.

The motivation behind this PR is to enable search result rendering on the server side as well. Previously, SSR would fail because the client cookie value wasn't being passed to the backend when running in authenticated mode.


frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
```diff
@@ -0,0 +1,49 @@
...
+  } else if (category?.startsWith(DocumentKind.Norm)) {
+    return `/v1/legislation`;
+  } else {
+    return `/v1/document`;
+  }
+}
+
+/**
+ * Converts the params used in the pinia store to URL params as expected by
+ * the API.
```
hmones, line 16: I think the function is self explanatory, it wouldn't need the comment. Also for the comments below. The code to me looks clean

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
```diff
@@ -0,0 +1,49 @@
+import type { QueryParams } from "~/stores/searchParams";
+import { DocumentKind } from "~/types";
+
+export function getUrl(category?: string) {
```
hmones, line 4: It might be a good idea to always add explicitly the return type to the function

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (category?.startsWith(DocumentKind.CaseLaw)) {
```
hmones, line 5: Isn't the category supposed to be exactly equal the value in the DocumentKind.CaseLaw? :thinking: using startsWith might be dangerous for maintainability of the code or for erroneous strings since the category string type is unchecked about (it's only string)

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+/**
+ * Converts the params used in the pinia store to URL params as expected by
+ * the API.
+ */
+export function convertParams(
+  params: Omit<QueryParams, "dateSearchMode"> & { temporalCoverage?: string },
+) {
```
hmones, line 20: Wouldn't we benefit here from a defined return type to the function? then we can know exactly how the returned json looks like, also the const result Record would have better typing rather than string types.

frontend/src/components/Search/SimpleSearch/SimpleSearch.vue
```diff
@@ -13,27 +12,33 @@ import { DocumentKind } from "@/types";
...
+import type { Page } from "~/components/Pagination/Pagination";
+import { convertParams, getUrl } from "./SimpleSearch.logic";
 
 const store = useSimpleSearchParamsStore();
 const values = storeToRefs(store);
+
+const config = useRuntimeConfig();
+const baseUrl = config.public.backendURL;
+
+const params = computed(() =>
```
hmones, line 24: If we have the type of converted Params we could use it here too.

hmones (commented):
Great work, I think this makes our code much better and maintainable since we rely on one library for the work :)

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
MasterCarl: Okay, fair point :)

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
MasterCarl: True, that would make things clearer indeed, and prevent bugs in case we want to refactor this.

frontend/src/components/Search/SimpleSearch/SimpleSearch.logic.ts
MasterCarl: I've added a comment here, this can be "R" or things like "R.Urteil". But we should consider moving all these to the enum, you are right.

PR #937, opened 2025-05-13
RISDEV-7642 Display legislation images

MasterCarl:
This PR adds support for displaying images and other files embedded in norms.

Image `src` attributes like `eli/bund/bgbl-1/2024/383/2024-12-19/1/deu/2024-12-19/illustration.png` will be converted to `/v1/legislation/eli/bund/bgbl-1/2024/383/2024-12-19/1/deu/2024-12-19/illustration.png` so that they can be loaded by the browser.

Since the internal portal will require authenticated requests to go through the Nuxt proxy, a header parameter `get-resources-via` is added. If set to PROXY, the URLs will be `/api/v1/legislation/eli/bund/bgbl-1/2024/383/2024-12-19/1/deu/2024-12-19/illustration.png`.

The XSLT has been adapted to support src and alt attributes, and convert the `akn:img` tag.


backend/src/main/java/de/bund/digitalservice/ris/search/utils/eli/ManifestationEli.java
```diff
@@ -20,7 +20,11 @@ public record ManifestationEli(
 
   @Override
   public String toString() {
```
hmones, line 22: The manifestation eli in principle should include the extension, therefore, I would suggest that we include in the builder of the manifestation eli the `.xml` and make it the default value if it wasn't set, but then we should pass other extension to the builder of the Manifestation eli since we would be working with more manifestations now for various urls.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/NormsController.java
```diff
@@ -411,8 +424,18 @@ public ResponseEntity<String> getLegislationArticleAsHtml(
               example = "hauptteil-1_art-1")
           @PathVariable
           String articleEid,
-      @PathVariable @Schema(allowableValues = "html") String format)
+      @PathVariable @Schema(allowableValues = "html") String format,
+      @RequestHeader(
+              name = ApiConfig.Headers.GET_RESOURCES_VIA,
```
hmones, line 429: Since this is a get operation, I would suggest to move this to a url parameter instead of a request header and make a validation on this url parameter with the enum values, we can also give it a default value if it is not set to be the non-api version. I think this would be a more compatible way with the REST get methods in my opinion.

backend/src/main/resources/XSLT/html/ldml_de/include/inhalt.xsl
```diff
@@ -592,6 +593,16 @@
         </dd>
     </xsl:template>
 
+    <!--
```
hmones, line 596: I'm not sure if this comment is very useful, we can remove it. It's understood below that this is an image node

hmones (commented):
I left some comments and suggestions. However, I believe that the way we fetch public xmls and resources would get out of hand at some point and can be simplified so much if we have a public bucket for public data where everything is referenced with their full url in the bucket that includes also the eli path. Just a food for thought, but I'll also write on the slack to think with the rest of the team.

backend/src/main/java/de/bund/digitalservice/ris/search/utils/eli/ManifestationEli.java
MasterCarl: That is a good point. I was also wondering whether a URI that ends with `.jpg` is a Manifestation ELI, or just something that looks like a Manifestation ELI. In the latter case, I would probably create a new class and revert the changes to `ManifestationEli.java`.

backend/src/main/resources/XSLT/html/ldml_de/include/inhalt.xsl
MasterCarl: I added this part for consistency with the existing style, which uses these block comments to organize the file into sections.

PR #953, opened 2025-05-26
Implement dedicated ZIP download

MasterCarl:
This PR implements a dedicated ZIP download button, in the Norm details tab.

Not having to distinguish between norms with and without attachments anymore, the logic could be simplified. The test cases have been updated accordingly.


hmones (changes requested):
Looks good to me, but can we still keep the data-attr `xml-zip-view` on the button so that it can be linked still to posthog download metrics?

PR #954, opened 2025-05-26
Add parsed request params to log json

MagdaN:


backend/src/main/java/de/bund/digitalservice/ris/search/utils/HttpLogSanitizer.java
```diff
@@ -1,21 +1,57 @@
...
     json.remove("origin");
     json.remove("host");
     json.remove("type");
 
+    json.put("queryParams", getQueryParamsAsText(request));
+
     return json;
   }
+
+  private static Map<String, String> getQueryParamsAsText(HttpRequest request) {
```
MasterCarl, line 31: I would call this getQueryParamsAsMap

backend/src/main/java/de/bund/digitalservice/ris/search/utils/HttpLogSanitizer.java
```diff
@@ -1,21 +1,57 @@
...
+import org.zalando.logbook.HttpRequest;
 
 public class HttpLogSanitizer {
 
   private HttpLogSanitizer() {
     throw new IllegalStateException("Utility class");
   }
 
-  public static JSONObject sanitizeLogJson(JSONObject json) {
+  public static JSONObject sanitizeLogJson(JSONObject json, HttpRequest request) {
```
MasterCarl, line 18: I think we should leave this method as-is and move `json.put("queryParams", ...)` to `LogRequestAndResponseTogetherSink::write` to keep the separation of concerns neat.

MasterCarl (changes requested):
Looks good, even though logging the query params as-is (without JSON processing) should also be fine, depending on our use case.

PR #956, opened 2025-05-30
RISDEV-7962

andrew-cenkner-digitalservice:
Bug with indexing initial condition
*Fixed bug

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexCaselawService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexCaselawService.java
@@ -97,6 +97,8 @@ public class IndexCaselawService implements IndexService {
   }
 
   private List<String> getAllCaseLawFilenames() {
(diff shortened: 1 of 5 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexCaselawService.java
```diff
@@ -97,6 +97,8 @@ public int getNumberOfFilesInBucket() {
   }
 
   private List<String> getAllCaseLawFilenames() {
-    return bucket.getAllKeys().stream().filter(s -> !s.contains(IndexSyncJob.CHANGELOG)).toList();
+    return bucket.getAllKeys().stream()
+        .filter(s -> !s.contains(IndexSyncJob.CHANGELOGS_PREFIX))
```
MasterCarl, line 101: A startsWith might be more straightforward here (but shouldn't matter)

MasterCarl (approved):
Looks good, thanks for fixing this!

PR #962, opened 2025-06-02
Close menu on click, enable active state

MasterCarl:
## Changes

- Keep "Menu" + icon appearance when the mobile menu is opened
- Close the mobile menu when an item is selected
- Add active page styling, using NuxtLink's native aria-current attribute selector
- Add test cases for the AppHeaderNav component


hmones (approved):
Good :)

PR #963, opened 2025-06-02
Implement basic Fassungen list

MasterCarl:
This PR implements a basic list of expressions for a given legislation work ("Fassungen").

A table layout was chosen for purported accessibility benefits.

The feature is hidden for the prototype environment.


hmones (changes requested):
Great work. I just think we need to add here an e2e test and this could be a chance to generate fake data for versioning that we can use in our testing later.  Also, we can specify on the tab to be loaded for either e2e or prototype, this way we can test it in the pipeline.

PR #973, opened 2025-06-11
RISDEV-8006 Page for historic versions of a norm

MasterCarl:
This PR makes navigation between different versions of a norm more convenient.

The main feature is a new breadcrumb element that links back to the "most relevant" version of a norm. It directs to a Nuxt route that performs a redirect,
- to the currently active version if available, or
- to a future or past version otherwise.
The last breadcrumb item shows the validity range of the currently open version, unless the open version is in force.

A few end-to-end cases and unit tests were added.


hmones:
@MasterCarl I updated the PR and merged the code from main to it (including the warning box) and will merge it to main so we have it on staging.

PR #987, opened 2025-06-17
Fix extraction of start and enddate of article

MagdaN:


backend/src/main/java/de/bund/digitalservice/ris/search/mapper/NormLdmlToOpenSearchMapper.java
```diff
@@ -432,8 +432,7 @@ private static LocalDate getArticleDate(
 
   private static String getEventReferenceEidByAttribute(
       Map<String, String> dateMap, String period) {
-    return dateMap.get(
-        "meta-1_geltzeiten-1_" + period.replace("#", "").trim() + "_gelzeitintervall-1");
+    return dateMap.get(period.replace("#", "").trim() + "_gelzeitintervall-1");
```
UrsKahmann, line 435: If I see this correctly `_gelzeitintervall-1` assumes that all articles always reference the first Geltungszeitinterval in the Geltungszeitgruppe?This is probably not safe as there can be multiple Geltungszeitintervalle inside a Geltungszeitgruppe and an article can reference any one of them. 
image

Example from a random file from the dump:
```
 <akn:article GUID="0541b296-71f8-4eee-a684-7e32efb80a5f"
                    eId="hauptteil-n1_buch-n14_art-n167"
                    period="#meta-n1_geltzeiten-n1_geltungszeitgr-n13">
```
(NOTE: Ignore the `n` prefix for the counters, they are introduced with LDML 1.8)

MagdaN: So if I understand correctly, I should be able to just just remove the `+ "_gelzeitintervall-1"` right?

PR #991, opened 2025-06-18
Revert change to CROSS_FIELDS multi match query, add test case

MasterCarl:
This PR is a re-creation and amendment of #984.

In the current implementation, searching for multiple terms doesn't match documents if

- the document contains these terms in different fields
  - where the fields are of different kinds (text, keyword).

This pull request adds a test case that covers this condition, and reverts the removal of query splitting that would cause the test to fail. It also renames the synthetic test document identifiers for easier identification.

If we want to keep the current implementation, an alternative way would be to add a secondary index type to all keyword fields:
```json
"field_name": {
  "type": "keyword",
  "fields": {
    "text": {
      "type": "text"
    }
  }
},
```


andrew-cenkner-digitalservice:
See discussion here https://digitalservicebund.slack.com/archives/C07EVJBF5K8/p1751291413132249

PR #992, opened 2025-06-18
Improve query documentation

MasterCarl:
This PR is a re-creation of #983.

This PR makes the code comments relating to OpenSearch more detailed, in order to illustrate certain design decisions.


hmones (approved):
I'm not generally in favor of verbose documentation, because it means the code isn't clear enough, but if you think this would be necessary to understand the code, let's add it then

PR #995, opened 2025-06-20
RISDEV-8036

andrew-cenkner-digitalservice:
Downloadable mock pdf (norm & caselaw)
*Refactored static strings
*Added new endpoint that returns a static pdf


MasterCarl (approved):
looks good to me!

PR #996, opened 2025-06-20
Fix and refactor date extraction for articles

MagdaN:


backend/src/main/java/de/bund/digitalservice/ris/search/models/ldml/TimeInterval.java
```diff
@@ -0,0 +1,9 @@
+package de.bund.digitalservice.ris.search.models.ldml;
+
+import lombok.Data;
+
+@Data
+public class TimeInterval {
```
MasterCarl, line 6: This might be a good use case for a `record`

backend/src/main/java/de/bund/digitalservice/ris/search/utils/LdmlTemporalData.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+public class LdmlTemporalData {
+
+  private static final Logger logger = LogManager.getLogger(LdmlTemporalData.class);
+
+  private static final String X_PATH_ALL_EVENT_REF = "//*[local-name()='eventRef']";
```
MasterCarl, line 22: While we're at it, could we refine this XPath to point to the actual node in the document instead of traversing the whole tree?

backend/src/test/java/de/bund/digitalservice/ris/search/unit/utils/LdmlTemporalDataTest.java
```diff
@@ -0,0 +1,96 @@
...
+         </akn:akomaNtoso>
+         """;
+
+  @Test
+  void testGetTemporalDataWithDatesMapping()
+      throws ParserConfigurationException, IOException, SAXException {
+    XmlDocument xmlDocument = new XmlDocument(testContent.getBytes());
+    Map<String, TimeInterval> result =
+        LdmlTemporalData.getTemporalDataWithDatesMapping(xmlDocument);
+    assert result.size() == 3;
```
MasterCarl, line 61: I would use AssertJ's `assertThat` methods here for better error messages

MasterCarl (commented):
I like that you extracted some code into LdmlTemporalData, which makes it slightly more readable. However, I'm not familiar enough with the LDML date logic to review this in substance.

backend/src/main/java/de/bund/digitalservice/ris/search/utils/LdmlTemporalData.java
```diff
@@ -0,0 +1,97 @@
...
+                    && timeIntervalNode.getNodeType() == Node.ELEMENT_NODE) {
+                  Element timeInterval = (Element) timeIntervalNode;
+                  TimeInterval timeIntervalData =
+                      fetchDatesForTimeInterval(timeInterval, eventRefMap);
+
+                  temporalDataWithDates.put(temporalGroupId, timeIntervalData);
+                }
+              });
+    } catch (XPathExpressionException e) {
+      logger.warn("Error parsing temporal data", e);
```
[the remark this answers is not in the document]
MagdaN: yes we should ignore it

backend/src/main/java/de/bund/digitalservice/ris/search/utils/LdmlTemporalData.java
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+    return temporalGroupToIntervalMap;
+  }
+
+  private static TimeInterval fetchDatesForTimeInterval(
+      Element timeIntervalNode, Map<String, String> eventRefMap) {
+    String start = eventRefMap.get(buildXpathFromID(timeIntervalNode.getAttribute("start")));
+    String end = eventRefMap.get(buildXpathFromID(timeIntervalNode.getAttribute("end")));
+    return new TimeInterval(start, (end != null && !end.isEmpty()) ? end : null);
```
andrew-cenkner-digitalservice, line 89: Is there a reason for the difference between start and end? I think either one (or both) can be null.

backend/src/main/java/de/bund/digitalservice/ris/search/utils/LdmlTemporalData.java
```diff
@@ -0,0 +1,126 @@
...
+
+    for (Node temporalGroup : temporalGroups) {
+      if (!(temporalGroup instanceof Element)) {
+        continue;
+      }
+
+      String temporalGroupId = String.format("#%s", getAttributeValue(temporalGroup, "eId"));
+      Node timeIntervalNode = getTimeIntervalNode((Element) temporalGroup);
+
+      if (timeIntervalNode != null && timeIntervalNode.getNodeType() == Node.ELEMENT_NODE) {
```
andrew-cenkner-digitalservice, line 76: Is this the same logic as line 69 (check if a node is an element)? If yes, is there a reason for the different approach?

andrew-cenkner-digitalservice (approved):

PR #1025, opened 2025-06-30
Add loading indicator

MasterCarl:
This enables the NuxtLoadingIndicator to indicate that a page is loading after a navigation action.

To see it in action, you may add a delay to `api/[...].ts`:

```ts
function delay(timeout: number = 1000): Promise<void> {
  return new Promise((resolve) => setTimeout(resolve, timeout));
}

export default defineEventHandler(async (event): Promise<unknown> => {
  await delay(1000);
  ...
}
```


UrsKahmann (approved):

hmones (approved):
That's a nice add :) Thanks.

PR #1029, opened 2025-07-01
Unify validityStatus calculation

UrsKahmann:
- use one function to calculate the validity status
- make sure time calculations happen in german local time
- rename ExpressionStatus to ValidityStatus as this better reflects what the status represents and also makes more sense when used with articles

### DIFF
--- a/frontend/src/utils/normUtils.ts
+++ b/frontend/src/utils/normUtils.ts
@@ -1,9 +1,10 @@
-import { formattedDate, formattedDateToDateTime } from "~/utils/dateFormatting";
+import { parseDateGermanLocalTime } from "~/utils/dateFormatting";
 import type { LegislationExpression, LegislationWork } from "~/types";
+import type { Dayjs } from "dayjs";
 
 export interface ValidityInterval {
-  from?: string;
-  to?: string;
+  from?: Dayjs;
+  to?: Dayjs;
 }
 
 export function temporalCoverageToValidityInterval(
@@ -11,7 +12,10 @@ export function temporalCoverageToValidityInterval(
 ): ValidityInterval | undefined {
   if (!temporalCoverage) return undefined;
   const [from, to] = temporalCoverage.replaceAll("..", "").split("/");
-  return { from: formattedDate(from), to: formattedDate(to) };
+  return {
+    from: parseDateGermanLocalTime(from),
+    to: parseDateGermanLocalTime(to),
+  };
 }
 
 export enum ExpressionStatus {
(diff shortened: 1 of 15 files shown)


frontend/src/utils/normUtils.ts
```diff
@@ -1,47 +1,55 @@
...
+    to: parseDateGermanLocalTime(to),
+  };
 }
 
-export enum ExpressionStatus {
+export enum ValidityStatus {
   InForce = "Aktuell gültig",
   Future = "Zukünftig in Kraft",
   Historical = "Außer Kraft",
 }
```
hmones, line 25: I think the naming here might not be very accurate because the status is the status of the whole entity/resource, which is in our case, either the Norm or the Expression of the norm. If we are very accurate, I would say it should be named the `NormExpressionStatus`, but we used a shortcut here and named it `ExpressionStatus`. If you are really against using the word Expression here, then at least we name it `NormStatus`. The keys also should change for semantic reasons and be something like: `FutureInForce` and `Expired` for future and historical Norms respectively. My suggestion would go into something like this (we can remove the Expression if you really dislike it even though it's very accurate to use here. 
```suggestion
export enum NormExpressionStatus {
  InForce = "Aktuell gültig",
  FutureInForce = "Zukünftig in Kraft",
  Expired = "Außer Kraft",
}
```

hmones (changes requested):
Thanks for offering to refactor this, it's greatly appreciated :heart: , I added some comments and I'm open for questions if you have any.

frontend/src/utils/normUtils.ts
```diff
@@ -1,47 +1,55 @@
...
+    (start.isBefore(currentDate, "day") || start.isSame(currentDate, "day")) &&
+    (end.isSame(currentDate, "day") || end.isAfter(currentDate, "day"))
+  ) {
+    return ValidityStatus.InForce;
+  } else if (
+    start.isAfter(currentDate, "day") &&
+    (end.isAfter(start, "day") || end.isSame(start, "day"))
+  ) {
+    return ValidityStatus.Future;
   }
```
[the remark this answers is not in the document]
UrsKahmann: Yes makes sense, I kinda wanted to be on the very safe side :D  but thats probably not necessary :)

frontend/src/utils/normUtils.ts
hmones: Thanks @UrsKahmann for the update.

Just to follow up on a couple of points:

You're right that in many cases it looks simpler and more convenient to use a unified Status enum, and we’re currently using it for both Norms and Articles (I had honestly forgotten that it's also used for Articles, since it's nested inside normUtils). However, from a semantic and structural point of view, the status of a Norm and that of a single article can carry different meanings and values. For example, the status "gegenstandslos" (void or obsolete) typically applies only to the Norm as a whole. We don’t handle this case yet, but we expect to in the future. So while having a global ValidityStatus may be a useful simplification for now, it's important to note that Norms and Articles may eventually require distinct statuses, and this should be kept in mind.

Regarding key naming: we had an internal discussion and agreed to stick with English variable and key names—especially since the project is open-source and intended for extensibility and external collaboration. Using keys like FutureInForce and Expired instead of German phrases like "Zukünftig in Kraft" improves clarity and maintainability, particularly for contributors who don’t speak German.

UrsKahmann: Okay so for point one: I would leave it like that for now (one status enum/union called `ValidityStatus` for articles and Norms) and if we have to add some other possible values (e.g. gegenstandslos etc) that don't make sense for both, then we have to adapt/split it in the future into (e.g. NormStatus and ArticleStatus)? I would vote against splitting them right now because maybe we will never need the differentiation #YAGNI Would that be fine with you?

Key Naming:
Okay I understand. And I agree maybe its better to decouple the status itself from how it it displayed. 
What I would suggest then is to change the enum back to a union type with "InForce" "FutureInForce" "Expired" instead of having an enum like:
```
 enum ValidityPeriod {
  InForce = "InForce",
  FutureInForce = "FutureInForce",
  Expired = "Expired",
}

```
which doesn't have any value as the keys and values are redundant. 
And then I would add another function like `getStatusLabel(status: ValidityStatus)` which will map the status to the german string to be used for displaying it. This way this function can be used in all places where the status is displayed and the display string stays consistent. Let me know if that would be okay with you? 

And thanks for the valuable feedback!

hmones: Great, just the name I would leave it ValidityStatus, not ValidityPeriod - Just English wise wouldn't be correct. Maybe you wrote it by msitake

UrsKahmann: Good catch sorry that was by mistake. It will of cause  stay ValidityStatus :)

UrsKahmann: > @UrsKahmann Today in the daily though another InForce status have been highlighted for Norms. Just taking that into consideration.

Yes, whenever it is decided how we deal with "gegenstandslos" and "bedingtes Inkrafttreten" we might need to adapt this then. I can take care of it :)

hmones:
> @hmones I pushed the changes as discussed, can you have a look again 🙏 thankssss

Looks great, let's merge :) :tada:

PR #1037, opened 2025-07-04
RISDEV-7608: Add spacing between list items

UrsKahmann:


hmones (commented):
Looks good to me. Just to mention, there's this class called `space-y-16` that can evenly space all child elements with 16 points instead of the mt on each element but other than that it's ok :) Thanks for mentioning me

UrsKahmann:
> Looks good to me. Just to mention, there's this class called `space-y-16` that can evenly space all child elements with 16 points instead of the mt on each element but other than that it's ok :) Thanks for mentioning me

Ahh thats so much nicer! Thanks for the hint 🫶

PR #1043, opened 2025-07-08
RISDEV-7643

andrew-cenkner-digitalservice:
Display inline images in caselaw
*Use new caselaw file structure in all environments


UrsKahmann (approved):
Looks okay 👍 I will integrate the change in [this PR](https://github.com/digitalservicebund/ris-search/pull/1034) then to reduce the code duplication :)

PR #1047, opened 2025-07-11
RISDEV-8529

andrew-cenkner-digitalservice:
Simple search | title search failing
*Refactor fields that are used as an identifier (ecli, eli, file number, etc.) to be indexed both as text and keyword. Text is used for the filtering logic just like the rest of the fields (every result must contain every token). Keyword is used for the targeted search boost ranking logic. An exact match (after normalization) of the entire search string gets a very large boost. *Fixed tests failing
*Refactored case law file number logic to work like the rest of the targeted search cases *Removed case law file number test that no longer makes sense *Fixed missing AND logic on search fields other than "searchTerm"

### DIFF
--- a/backend/src/test/java/de/bund/digitalservice/ris/search/integration/controller/api/AdvancedSearchControllerApiTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/search/integration/controller/api/AdvancedSearchControllerApiTest.java
@@ -4,7 +4,6 @@ import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.hasSize;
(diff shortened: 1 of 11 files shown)


backend/src/test/java/de/bund/digitalservice/ris/search/integration/controller/api/AdvancedSearchControllerApiTest.java
```diff
@@ -61,7 +60,7 @@ void setUpSearchControllerApiTest() throws IOException {
     if (initialized) return; // replacement for @BeforeAll setup, which causes errors
     initialized = true;
 
-    assertTrue(openSearchContainer.isRunning());
+    Assertions.assertTrue(openSearchContainer.isRunning());
```
UrsKahmann, line 63: **question:** This was probably automatically added by the IDE? I don't see the need for the `Assertions.`.

PR #1057, opened 2025-07-14
RISDEV-8569: Incorporate ldml.de 1.8.1 changes into test data

UrsKahmann:
This PR:
- updates the ldml.de test data files for e2e tests and integration tests to be schema conform to ldml.de 1.8.1
- adapts integrationtests/pact tests/e2e tests to work with the changes
- adds a test that runs schema validation on our test data files (this ensures that these files stay schema conform and our tests work with valid (as of schema conform) ldml.de test data) 

This PR:
- does NOT change the data in
        - `backend/src/test/resources/data/xmlTests`
        - `backend/src/test/resources/data/XsltTransformerServiceTest`
- these files are rather excerpts of ldml files (not complete ldml files), updating them can only be done in a best effort way because schema validating them is not really possible / makes no sense as attributes like `GUID` are omitted for simplicity. 
- I will look into how to deal with these files separately to this PR


hmones (commented):
It looks great to me, thanks for taking the initiative to do this :)

PR #1060, opened 2025-07-16
RISDEV-5478 Display Norms PDF links

hmones:
This PR implements the display of links for PDF files according to the figma designs attached in the ticket It also implements a limitation for getting files with extesions that are not supported so that the endpoint for retrieving files wouldn't be misused.


UrsKahmann:
> > Looking good 🚀 Only thing I was wondering about: The PDF endpoint you removed was theoretically for Manifestations in PDF format right? And not for attachments? So the endpoint for resources now supports pdf resources (attachments in pdf format) but if we get Manifestation PDFs (generated from the other streams) we'll have to add the PDF endpoint again right?
> 
> That's a good question, I was discussing the same thing with Andrew earlier, that the new merged endpoint would in fact serve all files that are supported whether they are PDF attachments or the main manifestation PDF, it will be served by the endpoint that is currently created. The change we have now is that if the manifestation PDF is requested it would return 404, which makes sense since we don't currently have the data migrated.

Thanks for the clarification, makes sense! But for me this opens the question: Why then keep a separate endpoint for the xml files and not serve them only with the "resource" endpoint as well? 🤔  
Maybe this is out of scope for the ticket and we should discuss it in the dev round :)

andrew-cenkner-digitalservice (approved):

PR #1064, opened 2025-07-18
RISDEV-8647 Reduce events sent to sentry in frontend and backend

ekl176:
This PR addresses the issue highlighted in the ticket [RISDEV-8647](https://digitalservicebund.atlassian.net/browse/RISDEV-8647) about the responsibility of the portal team to ~60% of the events in sentry for the company. 

- **Acceptance Criteria**
We expect to send ~10K events per month or less

[RISDEV-8647]: https://digitalservicebund.atlassian.net/browse/RISDEV-8647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/config/SentryConfig.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/config/SentryConfig.java
@@ -1,7 +1,9 @@
 package de.bund.digitalservice.ris.search.config;
 
+import io.sentry.SentryLevel;
 import io.sentry.SentryOptions;
 import jakarta.servlet.http.HttpServletRequest;
+import java.util.List;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
@@ -20,6 +22,16 @@ public class SentryConfig {
       } else {
         return null;
       }
+    };
+  }
+
+  @Bean
+  SentryOptions.BeforeSendCallback beforeSendCallback() {
+    return (event, hint) -> {
+      if (List.of(SentryLevel.ERROR, SentryLevel.FATAL).contains(event.getLevel())) {
+        return event;
+      }
+      return null;
     };
   }
 }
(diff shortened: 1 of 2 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/config/SentryConfig.java
```diff
@@ -22,4 +24,14 @@ SentryOptions.TracesSamplerCallback tracesSamplerCallback() {
...
+  @Bean
+  SentryOptions.BeforeSendCallback beforeSendCallback() {
+    return (event, hint) -> {
+      if (List.of(SentryLevel.ERROR, SentryLevel.FATAL).contains(event.getLevel())) {
+        return event;
+      }
+      return null;
+    };
+  }
 }
```
UrsKahmann, line 37: **question**: I wonder if we should set the [sampleRate](https://docs.sentry.io/platforms/java/configuration/sampling/#sampling-error-events) here, too. Because the default is 1.0 - would probably make sense to explicitly set it to 0.1?

UrsKahmann: Also in the `tracesSamplerCallback()` callback, should we maybe also return 0.1 intstead of null as "default"?

frontend/sentry.server.config.ts
```diff
@@ -15,8 +15,10 @@ Sentry.init({
   // It has to be in the environment of the server when building the app
   // See: https://nuxt.com/docs/guide/directory-structure/env#production
   dsn,
-  tracesSampleRate: 1.0,
+  tracesSampleRate: 0.1,
```
UrsKahmann, line 18: **question**: Should we also set the `sampleRate` to 0.1? Or is the `tracesSampleRate` sufficient?

PR #1083, opened 2025-07-30
RISDEV-8763 Create sitemap for portal static pages

hmones:
This PR implements a sitemap.xml for the static pages of the portal It utilizes the SEO NuxtJS modules and more specifically the sitemap module in order to build the sitemap dynamically.

Dynamic norms and caselaw pages are leftout of the scope of this PR as they will be implemented separately and would be done at indexing time.

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -1,10 +1,11 @@
 import tailwindcss from "@tailwindcss/vite";
+import { parseInt } from "lodash";
 import { defineNuxtConfig } from "nuxt/config";
 import IconsResolver from "unplugin-icons/resolver";
 import Icons from "unplugin-icons/vite";
 import { PrimeVueResolver } from "unplugin-vue-components/resolvers";
 import Components from "unplugin-vue-components/vite";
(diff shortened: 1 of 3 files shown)


UrsKahmann:
Nice 🚀  Do you think it would make sense to add an e2e test that makes sure that the sitemap is generated (correctly)?

hmones:
> Nice 🚀 Do you think it would make sense to add an e2e test that makes sure that the sitemap is generated (correctly)?

I was thinking to include tests from Lighthouse library at the end (indicated by a separate ticket) to check all SEO features and it also makes sure that the sitemap index is covered correctly. Now we have only the static pages.

UrsKahmann:
> > Nice 🚀 Do you think it would make sense to add an e2e test that makes sure that the sitemap is generated (correctly)?
> 
> I was thinking to include tests from Lighthouse library at the end (indicated by a separate ticket) to check all SEO features and it also makes sure that the sitemap index is covered correctly. Now we have only the static pages.

Sounds great 👍  Go for merge then from my side 🚀

PR #1108, opened 2025-08-15
add specific sort schemas per doctype

ekl176:
This PR sets up Sort Parameter schemas for every doctype instead of relying on a shared one.

This enables us to validate the sort values before a request hits the Controller and add api descriptions that reflects the actual possibilities of each endpoint.


backend/src/main/java/de/bund/digitalservice/ris/search/mapper/SortParamsConverter.java
```diff
@@ -0,0 +1,33 @@
+package de.bund.digitalservice.ris.search.mapper;
+
+import lombok.NonNull;
+import org.apache.logging.log4j.util.Strings;
+import org.springframework.data.domain.Sort;
+
+public class SortParamsConverter {
```
andrew-cenkner-digitalservice, line 7: Did you consider refactoring to have a class that implements `Converter` like done here https://www.baeldung.com/spring-mvc-custom-data-binder ?
The converter would take a string and return an Enum. The enum would list all possible sort values and have a boolean field for sort direction.
The first line of the controller could then do an Enum validation check instead of the current regex style and if it's not an allowed sort type for that controller, it would either return an error or just set it to the default (relevance).

ekl176: This class a mainly a renaming of the PaginationParamsConverter. I wanted to keep the change as incremental as possible. 
We'd need multiple Enums and Converters since they differ between endpoints.

PR #1111, opened 2025-08-19
RISDEV-8857

andrew-cenkner-digitalservice:
Switch from cross fields to pre tokenization
* Refactor UniversalDocumentQueryBuilder into 3 parts. A shared part, a norm part and a case law part. 
* Refactor how quoted and unquoted sections are parsed from searchTerm. Now if there is an unmatched ", the last one is removed before parsing. This simplifies the parsing logic. 
* The exact way that PortalQueryBuilder keeps it's state until the query is built is a potential topic for a future refactor. 
* The code that was moved to CaseLawQueryBuilder and NormQueryBuilder look to have some bugs and are a topic for future work.


backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/CaseLawQueryBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+        // use a match query to get subtypes, e.g., "Teilurteil" for query "Urteil"
+        boolQuery.should(matchQuery(DOCUMENT_TYPE, group.toString().toLowerCase()));
+      }
+    }
+    query.filter(boolQuery);
+  }
+
+  private static void queryDocumentType(@NotNull String[] types, BoolQueryBuilder query) {
+    // use the document_type.keyword field to match the query exactly
+    Stream<TermQueryBuilder> termQueries =
```
UrsKahmann, line 74: **hint:** This could be simplified like this:
```
    var boolQuery = QueryBuilders.boolQuery().minimumShouldMatch(1);
    Arrays.stream(types)
            .map(documentType -> QueryBuilders.termQuery(DOCUMENT_TYPE + ".keyword", documentType))
            .forEach(boolQuery::should);
    query.must(boolQuery);
```

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/NormQueryBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+  private NormQueryBuilder() {}
+
+  static final int ARTICLE_INNER_HITS_SIZE = 3;
+
+  public static void addSearchTerm(
+      String searchTerm,
+      List<String> unquotedSearchTerms,
+      List<String> quotedSearchPhrases,
+      BoolQueryBuilder query) {
+    // --- Article Search within Nested Documents ---
```
UrsKahmann, line 32: **praise:** I really like the comments here, they help a lot to understand the code!

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/NormQueryBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+    articleQueryBuilder.innerHit(articleInnerHitBuilder);
+
+    query.should(articleQueryBuilder);
+  }
+
+  public static void addNormFilters(NormsSearchParams params, BoolQueryBuilder query) {
+    if (params == null) {
+      return;
+    }
+    if (params.getEli() != null) {
```
UrsKahmann, line 95: **suggestion:** Use `Optional.ofNullable(..).ifPresent(...)` and the constant from `Norm.Fields` for "work_eli".

backend/src/test/java/de/bund/digitalservice/ris/search/unit/utils/QuotedStringParserTest.java
```diff
@@ -3,113 +3,101 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import de.bund.digitalservice.ris.search.utils.QuotedStringParser;
+import de.bund.digitalservice.ris.search.service.helper.QuotableSearchTerm;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 
 class QuotedStringParserTest {
```
UrsKahmann, line 10: **suggestion:** Rename this to `QuotableSearchTermTest` (also the filename) so it matches the record under test.

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/QuotableSearchTerm.java
```diff
@@ -0,0 +1,36 @@
...
+
+public record QuotableSearchTerm(
+    String original, List<String> unquotedSearchTerms, List<String> quotedSearchPhrases) {
+
+  public static QuotableSearchTerm parse(String searchTerm) {
+    List<String> unquotedSearchTerms = new ArrayList<>();
+    List<String> quotedSearchPhrases = new ArrayList<>();
+
+    // If " is unbalanced remove the last one
+    if (StringUtils.countMatches(searchTerm, "\"") % 2 == 1) {
```
UrsKahmann, line 15: **question:** Is this actually wanted in all cases?

I was wondering for example if a norm contains something like `.... they should be called "best programmer in the world".` Then I would expect to get an exact match when I search for `"in the world"."` right? But this won't be possible with this kind of hidden (to the user) transformation? 
(I know this is a superficial example, let me know if it doesn't make sense at all)

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/QuotableSearchTerm.java
```diff
@@ -0,0 +1,36 @@
+package de.bund.digitalservice.ris.search.service.helper;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang3.StringUtils;
+
+public record QuotableSearchTerm(
+    String original, List<String> unquotedSearchTerms, List<String> quotedSearchPhrases) {
+
+  public static QuotableSearchTerm parse(String searchTerm) {
```
UrsKahmann, line 10: **thought:** It would be helpful to have a short doc comment as to what this function does (maybe with a small example) as its rather complicated (in my opinion) to grasp what it does.

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/CaseLawQueryBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+import de.bund.digitalservice.ris.search.models.opensearch.CaseLawDocumentationUnit;
+import java.util.Arrays;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.Operator;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.index.query.TermQueryBuilder;
+
+public class CaseLawQueryBuilder {
```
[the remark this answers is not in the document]
andrew-cenkner-digitalservice: These lines were not added, just moved. I'll keep the suggestions in mind for my next PR making another refactor to the same code.

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/CaseLawQueryBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final String DOCUMENT_TYPE = "document_type";
+
+  private CaseLawQueryBuilder() {}
+
+  public static void addCaseLawFilters(CaseLawSearchParams params, BoolQueryBuilder query) {
+    if (params == null) {
+      return;
+    }
+    if (params.getEcli() != null) {
+      query.must(matchQuery("ecli", params.getEcli()).operator(Operator.AND));
```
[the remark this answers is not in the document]
andrew-cenkner-digitalservice: These lines were not added, just moved. I'll keep the suggestions in mind for my next PR making another refactor to the same code.

backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/QuotableSearchTerm.java
andrew-cenkner-digitalservice: I'll add a comment in the next PR.

UrsKahmann (approved):

PR #1112, opened 2025-08-19
RISDEV-8761 Integrate docs endpoints in portal

hmones:
This PR implements the integration of the endpoints documentation inside the portal.

What is not covered:
Links to other static pages.


MagdaN (commented):
I think this interface looks better then what we already have, only downside would be that the demo feature is not responsive.

PR #1115, opened 2025-08-21
RISDEV-8857

andrew-cenkner-digitalservice:
Switch from cross fields to pre tokenization
* Refactor the way searchTerm is processed
* Call opensearch with the unquoted parts of searchTerm to get the exact tokenization opensearch will use 
* Switch to BEST_FIELDS
* Other minor refactors


backend/src/main/java/de/bund/digitalservice/ris/search/service/helper/CaseLawQueryBuilder.java
UrsKahmann, line 1: I like it 🚀

backend/src/main/java/de/bund/digitalservice/ris/search/service/SearchTermService.java
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      return input;
+    }
+  }
+
+  private List<String> tokenize(String textToTokenize) {
+    // our OTC setup doesn't allow a global analyzer definition, but we use multiple copies of the
+    // exact same analyzer. Therefore,  it doesn't matter which index we call.
+    AnalyzeRequest request =
+        AnalyzeRequest.withIndexAnalyzer("norms", "custom_german_analyzer", textToTokenize);
```
UrsKahmann, line 70: We need to keep this in mind, if for some reason the analyzers diverge (e.g. we change something only for the norms analyzer) this could cause a hard to find bug right? 🤔 🐞

backend/src/test/java/de/bund/digitalservice/ris/search/integration/service/SearchTermServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+    ParsedSearchTerm result = searchTermService.parse("hello world");
+
+    assertEquals(List.of("hello", "world"), result.unquotedTokens());
+    assertTrue(result.quotedSearchPhrases().isEmpty());
+  }
+
+  @Test
+  void testOnlyQuotedTerms() {
+    ParsedSearchTerm result = searchTermService.parse("\"hello world\"");
+    assertTrue(result.unquotedTokens().isEmpty());
```
UrsKahmann, line 39: More of a note (maybe worth a topic in the dev round), I'd prefer the more verbose/readable `assertThat(...)...` syntax from the `AssertJ` liberary. 
E.g.
`assertThat(result.unquotedTokens()).isEmpty();`
or 
`assertThat(result.quotedSearchPhrases()).containsExactly("hello world")`
(I know these are just the tests from the former QuotedStringParser and they were already like that)

UrsKahmann (approved):
Looks good to me 🚀  Left some notes but nothing blocking :)

PR #1165, opened 2025-09-17
fix(a11y): add aria-label to dropdown RISDEV-9346

hamo225:
- Updated RIS UI to include accessible name on the icon-only dropdown button

-Addresses axe rule button-name on Caselaw Search page

RISDEV-9346


hmones (approved):
Perfect.
Just let's make sure the version bump from 11 to 14 didn't include any change to our used components elsewhere (visually I mean). Otherwise, let's merge it :)

PR #1179, opened 2025-09-23
RISDEV-9382 adding mobile as a separate config

hmones:
This PR improves the testing strategy by adding mobile as a separate project and making the matrix for e2e testing more verbose


frontend/e2e/login.public.spec.ts
```diff
@@ -1,11 +1,11 @@
...
 const displayName = process.env.E2E_KEYCLOAK_USER_DISPLAY_NAME ?? "Jane Doe";
 
-test("is authenticated", async ({ page }) => {
-  // authentication should happen in the setup flow in auth.setup.ts
+test("is authenticated", async ({ page }, workerInfo) => {
   await page.goto("/");
-
-  const expectedLogoutButtonName = `${displayName} Abmelden`;
-  await expect(page.getByText(expectedLogoutButtonName)).toBeVisible();
+  if (workerInfo.project.name === "mobile")
```
andreasphil, line 7: todo: instead of repeating this condition and hardcoding the project name in every relevant test, IMO there should be a [fixture](https://playwright.dev/docs/test-fixtures) called `isMobileTest`:

```ts
test("is authenticated", async ({page, isMobileTest}) => {
  // ...
})
```

frontend/e2e/normsView.spec.ts
```diff
@@ -203,90 +203,96 @@ test("can view images", async ({ page }) => {
...
+  for (const testCase of testCases) {
+    test(
+      testCase.name,
+      async ({ page, browserName, baseURL, context }, workerInfo) => {
+        const isMobileTest = workerInfo.project.name === "mobile";
+        await page.goto("/norms/eli/bund/bgbl-1/2024/383/2024-12-19/1/deu", {
+          waitUntil: "networkidle",
+        });
+
+        if (browserName === "chromium") {
```
andreasphil, line 227: question: why do we need Chrome-specific logic here?

frontend/playwright.config.ts
```diff
@@ -47,6 +48,16 @@ const browserConfigurations: Project[] = [
     },
     dependencies: ["setup"],
   },
+  {
+    name: "mobile",
+    use: {
+      ...devices["Desktop Firefox"],
```
hamo225, line 54: Can’t remember what we discussed. We’re using "Desktop Firefox" with a small viewport here instead of an actual mobile device, right? whats the best practice here?  I assume it works but just curious.

hmones: We can get an actual device, but we wanted to not use the name of a device to mislead people that we are testing on a specific device name, so I made a custom configuration here for our own extending the desktop of firefox and then enabling touch (disabling hover ...etc) and limiting the viewport to certain dimensions

hamo225: Ok yeah i get it. Covering the basics with this custom setup, so does the job we need without misleading with the device name 👍. I wonder if we may miss a few things out of the box when not using a built-in mobile device? Maybe something to look into later, but i also guess this works fine for the current needs.

PR #1185, opened 2025-09-29
Refactor xslt transformers

UrsKahmann:
- in preparation for implementing the literature (and later VV) xslt transformation the xslt transformer is split to prevent having on large transformer class in the end which is handling all document kinds


VictorDelCampo (commented):
Hey @UrsKahmann LGTM, just two minor comments

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/LiteratureController.java
```diff
@@ -93,17 +89,9 @@ public ResponseEntity<String> getLiteratureAsHtml(
...
-
-    if (bytes.isPresent()) {
-      String html = xsltTransformerService.transformCaseLaw(bytes.get(), resourcePath);
-      return ResponseEntity.ok(html);
-    } else {
-      return ResponseEntity.notFound().build();
-    }
+          ResourceReferenceMode resourceReferenceMode) {
+    // implement this correctly when the literature xslt transformer is available
+    return ResponseEntity.notFound().build();
```
VictorDelCampo, line 94: It is big nitpickying but I would throw a not-implemented status exception instead

`throw new ResponseStatusException(HttpStatus.NOT_IMPLEMENTED, "Not implemented yet");`

Didn't check though is there is a controller advice transforming unexpected errors into something unrecognizable

PR #1191, opened 2025-09-30
Date filter & advanced search page integration

andreasphil:
This PR adds the date filter and a first version of the overall integration of the advanced search page. Even though it is not done yet, I would like to merge this soon to prevent this PR from growing to large and becoming unreviewable.

Not yet implemented:

- Unit tests for advanced search composable
- Badge indicating "in force" status of a norm for norm search results
- E2E Tests

Other known issues:

- I want to remove the logic for storing/loading the filters from the URL outside the view to make it easier to test
- The extra classes for the panel menu shouldn't be hardcoded there, will think about a better solution
- The way the URL is constructed for the API calls is probably broken though I don't know yet how exactly it will fail :D will need to check once deployed
- If you sort by some criterion that is only available for one document kind and switch document kinds, the filter breaks
- Responsive: the filter bar is broken on mobile + I'm not sure about the order of elements on the page on mobile
- Shows an empty result list before the initial search
- The norms filter should default to "Aktuell gültig"

Out of scope / won't fix:

- The date filter doesn't do anything ([ticket](https://digitalservicebund.atlassian.net/browse/RISDEV-9075))
- Showing the total count of data sets ([ticket](https://digitalservicebund.atlassian.net/browse/RISDEV-9488))
- Norms results not showing highlights (see discussion on [Slack](https://digitalservicebund.slack.com/archives/C07EVJBF5K8/p1759244164429109))
- Help box for operators ([ticket](https://digitalservicebund.atlassian.net/browse/RISDEV-9204))
- Cleanup of old search code ([ticket](https://digitalservicebund.atlassian.net/browse/RISDEV-8504))
- The data fields that are inserted when clicking something don't always match the actual field names as they're currently implemented. This is intended. Reach out to me or Nadine if you have questions.

Anything not listed here: please let me know :)

RISDEV-9205, RISDEV-8385

### DIFF
--- a/frontend/src/composables/useAdvancedSearch.ts
+++ b/frontend/src/composables/useAdvancedSearch.ts
@@ -7,7 +7,7 @@ import type { Page } from "~/components/Pagination/Pagination";
 import { DocumentKind } from "~/types";
 
 /** Additional configuration for search API calls */
-type UseAdvancedSearchOptions = {
+type AdvancedSearchOptions = {
   /** Number of search results per page */
   itemsPerPage: MaybeRefOrGetter<string>;
 
@@ -35,7 +35,7 @@ export async function useAdvancedSearch(
     itemsPerPage = "50",
     pageIndex = 0,
     sort = "default",
-  }: Partial<UseAdvancedSearchOptions>,
+  }: Partial<AdvancedSearchOptions>,
 ) {
   const searchEndpointUrl = computed(() => {
     const documentKindVal = toValue(documentKind);


frontend/src/utils/documentKind.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { isDocumentKind } from "./documentKind";
+
+describe("documentKind", () => {
+  describe("isDocumentKind", () => {
+    it("returns true if the value is a known document kind", () => {
+      for (const i of ["A", "R", "N"]) {
```
UrsKahmann, line 6: **suggestion:** I know this is not really part of the PR but reading this I had a hard time figuring out what "R" is. After looking into the `DocumentKind` enum I think it would be more understandable is Caselaw mapps to "C".

But I know this goes into the topic of English vs German namings which we had discussions about but no super clear rules. Just wanted point it out as I was confused by the "R" :D

frontend/src/components/Ris/RisDateInput.vue
```diff
@@ -0,0 +1,187 @@
...
+      class="w-full"
+      mask="99.99.9999"
+      placeholder="TT.MM.JJJJ"
+      @blur="onBlur"
+      @keydown="backspaceDelete"
+    />
+
+    <small v-if="localValidationError" :id="`${id}-hint`">
+      <IconErrorOutline />
+      {{ errorMessage }}
```
UrsKahmann, line 184: **question:** Not sure if I am doing something wrong or its just not supposed to show yet, but I couldn't get this error message to show: 
image
Is that intended?

UrsKahmann (approved):
Thanks you for the nice work 🚀  I left a few comments but nothing blocking, some are also just questions to get a better understanding of certain constructs. Feel free to merge and address the comments (afterwards) and only if you feel like they make sense ✌️

frontend/src/components/Ris/RisDateInput.vue
andreasphil: Good catch, fixed it!

PR #1200, opened 2025-10-02
Risdev 8450 simple search implement stakeholder prioritization

andrew-cenkner-digitalservice:
* Added Map of boost values based on specification
* Added logic to use boost maps at query time
* Removed boosts added in mapping files
* Removed some dead code
* Updated some names for clarity
* Split the analyzer into two copies. One used for the document_type with decompounding and one for the rest of the fields so the decompounding problem doesn't impact the rest of the index. This is possible since we no longer use CROSS_FIELDS
*Refactored simple search query generation
* Refactor SimpleSearchQueryBuilder moving stream specific logic to the respective Type class
* Fix merge conflict
* Refactor Literature simple search to match abstracted style
* Fix failing tests


backend/src/main/java/de/bund/digitalservice/ris/search/service/AllDocumentsService.java
```diff
@@ -58,14 +54,18 @@ public AllDocumentsService(
    * @param pageable Page (offset) and size parameters.
    * @return A new {@link SearchPage} of the containing {@link AbstractSearchEntity}.
    */
-  public SearchPage<AbstractSearchEntity> searchAllDocuments(
-      final String search, Pageable pageable) {
+  public SearchPage<AbstractSearchEntity> advancedSearchAllDocuments(
+      String search, Pageable pageable) {
+    HighlightBuilder highlightBuilder = RisHighlightBuilder.baseHighlighter();
+    new NormSimpleSearchType(null).addHighlightedFields(highlightBuilder);
```
UrsKahmann, line 60: **suggestion:** Revert this call so the `addFields` is called on the `highlightBuilder`.

Reading this is really confusing because it reads like there will be highlighted fields added to the search type but then a highlight builder is passed as argument. In my opinion things would be a lot clearer if the call would be something like:
`highlightBuilder.addFields(NormSimpleSearchType.NORMS_HIGHLIGHT_CONTENT_FIELDS)` 

(same for the other search types)

UrsKahmann (approved):
Nice work! 🚀  I only added one comment but nothing blocking. I really like the refactoring and that the query building code is removed from the services. Also the introduction of the SearchType classes seems logical. But I have to admit: I still don't grasp the whole query building logic 100% 😅

PR #1838, opened 2026-04-08
RISDEV-11395

andrew-cenkner-digitalservice:
Backend: Fix UserDetailsServiceAutoConfiguration
*Lower docker image requirements so it fits in default levels
*Configure spring boot to not setup a default user which we don't need


UrsKahmann (approved):
Thanks for fixing this 🚀

PR #1851, opened 2026-04-13
RISDEV-0000 use default PNPM linking

andreasphil:
When we introduced PNPM as our package manager, we [set](https://pnpm.io/settings#nodelinker) the `nodeLinker` to `hoisted`, which results in the same `node_modules` structure as NPM. This was done for compatibility reasons; otherwise our application would fail to run. The main reason for this is that PNPM by default doesn't allow application code to access dependencies that aren't explicitly declared in the project's `package.json` (in our case: Vite and Vue, managed by Nuxt).

However, PNPM by default uses a different structure for good reasons (performance and safety, more information [here](https://pnpm.io/motivation#creating-a-non-flat-node_modules-directory)). This PR:

- Switches to PNPM's default behavior of using symbolic linking for dependencies
- Manually hoists `vue` and `vite`---we want Nuxt to continue to manage these for us instead of declaring them explicitly. This gives our application code access to them without having to declare them as an explicit dependency.
- Removes the Sentry test page as it was accessing an internal API, now no longer allowed (but we don't need it for now)

> [!NOTE]
>
> You may need to `rm -rf node_modules && pnpm install` to ensure the changes are applied.


UrsKahmann (approved):
Thanks for taking care of that 🚀

PR #1853, opened 2026-04-13
RISDEV-0000 fix broken escaping of commit messages in CI

andreasphil:
When committing updated OpenAPI specs, escaping the commit message with `toJSON` isn't sufficient because it doesn't escape backticks. Those are often used in Markdown-formatted commit messages, causing the pipeline to fail. Since backticks may also be interpreted as command substitution in shell environments, it additionally exposes us to shell injection attacks.

See
https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -271,11 +271,13 @@ jobs:
           path: docs/data/
           name: openapi.json
       - name: Push reports
+        env:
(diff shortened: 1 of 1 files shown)


UrsKahmann:
Thanks for fixing this 🙏

PR #1856, opened 2026-04-14
RISDEV-11413 simplify eID encoding, decoding and Umlaut conversion

andreasphil:
This PR:

- **Removes all logic for converting Umlaute in eIDs:** We used to translate Umlaute in eIDs to their `ae`, `oe`, `ue` equivalents. It's a constant source of bugs and confusing behavior and is better handled by the browser, frontend router, and Spring endpoints. In addition, LDML.de 1.9 will change the specification to remove Umlaute from eIDs altogether (though if and when we upgrade is not clear at this point).

- **Further simplifies URL encoding/decoding in the frontend** by delegating URL construction to the router where possible, letting it handle this stuff for us; clearly documents edge cases where manual encoding and decoding is still necessary.

- Adds E2E tests for previously uncovered and potentially tricky eIDs

### DIFF
--- a/backend/src/main/resources/XSLT/html/case-law.xslt
+++ b/backend/src/main/resources/XSLT/html/case-law.xslt
@@ -8,14 +8,6 @@
 
 	<xsl:output method="html" encoding="UTF-8" />
 
-	<xsl:function name="local:encode-for-uri" as="xs:string">
-		<xsl:param name="uri" as="xs:string" />
-		<xsl:variable name="uri" select="replace($uri, 'ä', 'ae')" />
-		<xsl:variable name="uri" select="replace($uri, 'ö', 'oe')" />
-		<xsl:variable name="uri" select="replace($uri, 'ü', 'ue')" />
-		<xsl:value-of select="$uri" />
-	</xsl:function>
-
 	<xsl:param name="ressourcenpfad" as="xs:string" select="''"/>
 
 	<!-- akn wrapper elements (can be removed but content is kept) -->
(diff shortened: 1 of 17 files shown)


frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue
Comment (UrsKahmann) [inline frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue:1]: **praise:** The comments you added are really helpful, thank you!

backend/src/main/resources/XSLT/html/case-law.xslt
UrsKahmann, line 1: **question:** This is the only file touching case-law related code. Apparently this "enoding" was not covered by a test so far (otherwise that would fail now I guess). I am not super sure how the border number linking work, but did you check that this doesn't break the linking?

UrsKahmann (approved):

UrsKahmann:
Really nice to get rid of this "encoding" mess! 🚀

backend/src/main/resources/XSLT/html/case-law.xslt
andreasphil: As far as I can tell this was just put there to be extra sure, but Randnummern (as the name suggests) should always be numbers, so the Umlaut conversion should not do anything here. In addition the border number itself (and its ID) are not encoded, so this would've been broken either way.

PR #1858, opened 2026-04-15
RISDEV-11430-consolidate-toc

ekl176:
This PR removes the tableOfContents property from the legislation schema. 

The structure of a legislation document will be reflected by the nested structure of the hasPart array of a document instead of using the custom tableOfContents. 
With this we reduce redundancy and stick closer to schema.org recommendations.

This PR has no impact to our opensearch mapping. We still store the structure (table of contents) independent from the content of the individual parts (articles array) of a document to stick to our query logic using nested queries for articles.


backend/src/test/java/de/bund/digitalservice/ris/search/unit/mapper/NormSchemaMapperTest.java
```diff
@@ -10,55 +10,13 @@
...
 import java.time.LocalDate;
 import java.util.List;
-import java.util.Objects;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 class NormSchemaMapperTest {
 
-  @Test
-  void itMapsATableOfContents() {
```
UrsKahmann, line 23: **suggestion:** Maybe add a similar test that makes sure given a norm with a nested TOC, the `hasParts` is created correctly? So that the changes in the `NormSchemaMapper` especially the `buildLegislationExpressionPart()` is covered.

frontend/src/composables/useNormData.ts
```diff
@@ -53,7 +53,6 @@ export function useFetchNormContent(
       const document = parseDocument(html);
       const htmlParts = extractHtmlParts(document);
       //set hasPart empty to reduce payload size
```
UrsKahmann: **todo:** Remove the comment as well.

UrsKahmann (approved):
Nice 🚀  I left two small comments ✌️

frontend/src/utils/tableOfContents.spec.ts
```diff
@@ -39,19 +42,23 @@ describe("tableOfContents", () => {
...
+          hasPart: [
             {
-              "@type": "TocEntry",
-              id: "art-1",
-              marker: "§ 1",
-              heading: "Scope",
-              children: [],
+              "@id": "art-1",
+              eId: "art-1",
+              headline: "§ 1",
```
andreasphil: nitpick: mixed up `name` and `headline` (headline should be "Scope" and name should be "§ 1")

frontend/src/composables/useNormData.ts
```diff
@@ -52,8 +52,6 @@ export function useFetchNormContent(
       });
       const document = parseDocument(html);
       const htmlParts = extractHtmlParts(document);
-      //set hasPart empty to reduce payload size
```
andreasphil, line 55: question: did you test what this means for large norms? I remember we specifically removed `hasPart` e.g. for the BGB, which significantly reduced payload size and improved performance

backend/src/main/java/de/bund/digitalservice/ris/search/mapper/NormSchemaMapper.java
```diff
@@ -95,35 +93,69 @@ public static List<LegislationObjectSchema> getEncodings(@Nullable String manife
...
-                    buildTableOfContents(item.children())))
+        .map(item -> buildLegislationExpressionPart(articles, item, idPrefix))
         .toList();
   }
 
-  private static List<LegislationExpressionPartSchema> buildPartList(Norm norm, String idPrefix) {
-    if (norm.getArticles() == null) return Collections.emptyList();
-    return norm.getArticles().stream().map(article -> buildPart(article, idPrefix)).toList();
+  /**
+   * Builds a single legislation expression part. It uses a combination of the tableOfCOntents and
```
andreasphil: typo: `tableOfCOntents`

```suggestion
   * Builds a single legislation expression part. It uses a combination of the tableOfContents and
```

frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue
```diff
@@ -39,9 +43,9 @@ if (error.value) {
 const normPath: string = route.fullPath.replace(/\/[^/]*$/, "");
 
 const tableOfContents = computed(() => {
-  if (!norm.value?.tableOfContents) return [];
+  if (!norm.value?.hasPart) return [];
```
Comment (andreasphil) [inline frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue:46]: suggestion: if I understand correctly it would be better to check `singleViewParts`, right? Because that seems to be where the navigation lives now. That should also make it unnecessary to use the `!` assertion below, that should be avoided either way

frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue
```diff
@@ -39,9 +43,9 @@ if (error.value) {
...
   return tocItemsToTreeViewItems(
-    norm.value.tableOfContents,
+    norm.value.hasPart!,
```
Comment (andreasphil) [inline frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue]: todo: check if you can avoid the `!` assertion. This should not be necessary given the guard above, if it still is that indicates some subtle bug in the logic

backend/src/main/java/de/bund/digitalservice/ris/search/schema/TableOfContentsSchema.java
```diff
@@ -15,6 +16,7 @@
  * with this entry. - children: A list of child entries forming a hierarchical structure under this
  * entry.
  */
+@Builder
 public record TableOfContentsSchema(
```
andreasphil: question: this schema shouldn't be needed anymore after the refactoring, correct?

andreasphil (changes requested):

frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue
Comment (ekl176) [reply frontend/src/pages/norms/eli/[jurisdiction]/[agent]/[year]/[naturalIdentifier]/[pointInTime]/[version]/[language]/[eId].vue:46]: singeViewParts only contains the leaf nodes of the tree to enable the navigation from one article to the next. 
To construct the tableOfContents we need the whole tree.

frontend/src/composables/useNormData.ts
ekl176: We were able to remove the hasPart because the tableOfContents made it redundant. With the table of contents gone we need the hasPart to receive the tree structure of the single parts of the document.

andreasphil (approved):

PR #1859, opened 2026-04-16
Add api-docs project

UrsKahmann:
- copy all files of the ris-portal-docs repo into `/api-docs`
- make sure dependabot considers `/api-dcos` for updates as well

RISDEV-11174


andreasphil (commented):
@UrsKahmann I got a heart attack from the amount of changed files 😅 but I assume it just copy & paste from the docs repo?

Good catch also updating Dependabot, but that brings me to my question: we currently have a [workflow](https://github.com/digitalservicebund/ris-search/blob/3c8bfc2105964708725dba9d52b46f2c082fa7a0/.github/workflows/backend.yml#L251) to push changes to the API spec to the docs repo, which we will no longer need, correct? So do we need to:

- remove that workflow? As long as the repo exists it's only a bit wasteful but I assume we want to delete that
- some form of automation to update the JSON in the docs folder when it changes?

UrsKahmann:
> @UrsKahmann I got a heart attack from the amount of changed files 😅 but I assume it just copy & paste from the docs repo?
> 
> Good catch also updating Dependabot, but that brings me to my question: we currently have a [workflow](https://github.com/digitalservicebund/ris-search/blob/3c8bfc2105964708725dba9d52b46f2c082fa7a0/.github/workflows/backend.yml#L251) to push changes to the API spec to the docs repo, which we will no longer need, correct? So do we need to:
> 
>     * remove that workflow? As long as the repo exists it's only a bit wasteful but I assume we want to delete that
> 
>     * some form of automation to update the JSON in the docs folder when it changes?

Sorryyy, yes its just a copy and paste. I should have given more context, my bad. As discussed a few weeks ago the general plan is to move a way from having the separate repo for the api docs and move them as separate application into the ris-search repo. Then deploy them as static site using a container image instead of using a bucket. 

So the the plan is:
1. first simply copy the files over (this PR) - but leave the docs repos and deployment (serving the files from a bucket) as it is for now
2. add a docker file and build step to the pipeline so the api docs are runnable as docker image
3. create a api docs deployment based on the image (and not the bucket anymore) with a temporary url
4. if that is running and works as expected -> switch the docs.testphase url to point to that new deployment, remove the pipeline steps that push to the docs repo and archive the docs repo
5. remove the endpoints files and link to swagger

PR #1864, opened 2026-04-17
RISDEV-11413 missing links in breadcrumb of Einzelnorm

andreasphil:
Fixes a regression in breadcrumbs of Einzelnormen introduced by da13f1f55b8041e7e9b3c6bae3d9ebc46dd691c5:

- all breadcrumbs now link to the correct position in the document again
- only the last element shows the full title, elements in between only appear labeled with their markers
- adds missing tests


UrsKahmann (commented):
Looks good! 🙌

PR #1865, opened 2026-04-17
RISDEV-10889 keep table of contents visible in short documents

andreasphil:


UrsKahmann:
This is so niceee 👌 🚀

PR #1891, opened 2026-04-21
Show custom 404 error page

UrsKahmann:
- adjust the nginx.config to show the applications 404 page instead of nginx's default error page

RISDEV-11175


andreasphil (commented):
👍

PR #1899, opened 2026-04-23
Add button linking to swagger

UrsKahmann:
- introduce link button that opens the swagger page in a new tab
- remove the "Endpoints" entry from the navigation sidebar
- there are more parts to the original endpoints implementation that should be removed as well, this will be done in a separate commit

### Open questions:
- the actual link to the swagger ui will only work because of the ingress configuration. Meaning:
    - docs.rechtsinformation.bund.de points to the api docs application
    - docs.rechtsinformation.bund.de/swagger-ui/** points to the backend application

I am not sure if this is a good appraoch?
Should the link rather be hardcoded (e.g. full url)? 

RISDEV-11297


api-docs/.vitepress/theme/components/ButtonLink.vue
```diff
@@ -1,10 +1,8 @@
 <script setup lang="ts">
-import { computed } from "vue";
-
 const props = defineProps<{ href: string }>();
 </script>
 <template>
-  <a :href="props.href" class="ds-button">
+  <a :href="props.href" v-bind="$attrs" class="ds-button">
```
andreasphil, line 5: question: why do we need the `v-bind` here? This should be covered automatically by [fallthrough attributes](https://vuejs.org/guide/components/attrs.html)

andreasphil:
> the actual link to the swagger ui will only work because of the ingress configuration.

I'm a bit undecided on that one, but I would tend to say: make it a full URL. Reason: readability - reading a URL like that would suggest to me that the route is somehow part of the same folder/project, but it isn't. Making it a full URL would make that clear. However for the full URL we would need some environment-specific configuration, correct? Unless we decide that it always points to the same public instance regardless of the environment. Not sure how complicated that is and if it's worth putting in that effort though...

api-docs/.vitepress/theme/components/ButtonLink.vue
UrsKahmann: True, will remove that :)

UrsKahmann:
> > the actual link to the swagger ui will only work because of the ingress configuration.
> 
> I'm a bit undecided on that one, but I would tend to say: make it a full URL. Reason: readability - reading a URL like that would suggest to me that the route is somehow part of the same folder/project, but it isn't. Making it a full URL would make that clear. However for the full URL we would need some environment-specific configuration, correct? Unless we decide that it always points to the same public instance regardless of the environment. Not sure how complicated that is and if it's worth putting in that effort though...

I agree the full url would be cleaner. I will check if there is an easy way to make the url configurable on a per delpoyment (envrionment) basis, if not I will just hardcode it for now as we only have one deployment. And then we can think about it again if we ever plan to have multiple deployments.

PR #1901, opened 2026-04-23
Fix button appearance in focus state

UrsKahmann:
- make sure the button is still readable in focus state
- hardcode the swagger url so its clear that this is no internal link
 
Ideally the url would be provided via a runtime configuration so it can be easily changed per deployment. But this doesn't seem super easy for a statically served site, so for simplicity and as we don't deploy the docs in multiple environments so far, a hardcoded url is used for now 

RISDEV-11297


andreasphil (commented):
👍

PR #1902, opened 2026-04-23
RISDEV-11270 new mobile design for TOC

andreasphil:
This PR:

- Reworks the (formerly Norms-only) table of contents to support the new mobile drawer UI pattern
- Shares the table of contents between all document types, applying the new mobile drawer pattern to all of them
- Updates unit and E2E tests


frontend/e2e/administrativeDirective.spec.ts
```diff
@@ -139,6 +139,52 @@ test("sidebar TOC is not shown when document has no text sections", async ({
...
+    await expect(page.getByRole("button", { name: "Inhalte" })).toBeVisible();
+  });
+
+  test("clicking the floating button opens the TOC", async ({
+    page,
+    isMobileTest,
+  }) => {
+    test.skip(!isMobileTest);
+    await navigate(page, "/administrative-directives/KSNR000000001");
+
```
UrsKahmann, line 159: **nitpick:** Maybe add a check before clicking, that makes sure that the TOC is not already visible? (But maybe that is to detailed 😅 )
(same for the others)
Or maybe it makes more sense to simply test that in the unit tests 🤔

UrsKahmann (commented):
Looking really nice! Just a tiny comment, ignore if to much :D

frontend/e2e/administrativeDirective.spec.ts
andreasphil: that's implicit in the click below, otherwise the click would not be working :)

PR #1908, opened 2026-04-24
RISDEV-11283 use drawer for mobile action menu

andreasphil:


frontend/src/components/documents/actionMenu/ActionMenu.vue
```diff
@@ -1,53 +1,96 @@
...
+      header="Aktionen"
+      position="bottom"
+      :close-button-props="{
+        size: 'small',
+        label: 'Schließen',
+        iconPos: 'right',
+      }"
+    >
+      <ul class="-mt-12">
+        <li v-for="item in actions">
```
UrsKahmann, line 55: **question:** I assume not using a specific key here at all now is okay because the items number and order never changes?

UrsKahmann (commented):
Awesome 🚀

frontend/src/components/documents/actionMenu/ActionMenu.vue
andreasphil: yes, I removed the key because we don't have a reliable identifier. But since the menu items are essentially hardcoded we don't need one in this case.

PR #1926, opened 2026-04-29
RISDEV-0000 tidy + optimize PrimeVue imports

andreasphil:
This PR refactors imports of PrimeVue components to consistently import from `primevue` instead of `primevue/component` wherever possible, and adds a linter rule to enforce the pattern for new code. This allows us to pre-bundle the PrimeVue package during development, which should dramatically reduce flickering and erratic page reloading due to dependency optimization.

I deliberately chose not to optimize _all_ possible dependencies, despite the Nuxt dev server suggesting so, since listing everything we need in the Vite config would be annoying to maintain 😅


ekl176 (commented):
✅

PR #1931, opened 2026-04-30
Align word breaking behaviour across document types

UrsKahmann:
RISDEV-9451

### DIFF
--- a/frontend/src/pages/translations/index.vue
+++ b/frontend/src/pages/translations/index.vue
@@ -124,6 +124,7 @@ const translationsListId = useId();
       >
         <div class="flex max-w-prose flex-col gap-8">
           <SearchResultHeader
+            lang="de"
             :items="[{ value: t['@id'] }, { value: t.translationOfWork ?? '' }]"
           />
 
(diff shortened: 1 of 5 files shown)


frontend/src/pages/translations/index.vue
```diff
@@ -124,6 +124,7 @@ const translationsListId = useId();
       >
         <div class="flex max-w-prose flex-col gap-8">
           <SearchResultHeader
+            lang="de"
```
andreasphil, line 127: question: what is this for?

andreasphil (commented):
@UrsKahmann looks good based on the code, but I will take some time at the beginning of next week to try and break it. Feel free to merge and move along with functional review in the meantime

frontend/src/pages/translations/index.vue
UrsKahmann: In the translations search results, the page has `lang="en"` but the search result headers are in german:
grafik
But for `hyphens-auto` to work, the correct language needs to be set. At least that was my understanding and that seemed to work. But let me know if I am on the wrong track :D 

EDIT: Maybe this is also a little bit out of scope because I was mostly trying to align the document pages, but noticed this and though I'd give it a try to fix it as well..

PR #1940, opened 2026-05-04
RISDEV-0000 upgrade PNPM

andreasphil:
- Upgrades PNPM to v11, fixes breaking changes
- Reduces places where the package manager is specified to `packageManager` property in `package.json`
- Aligns config of api-docs package with frontend


UrsKahmann (commented):
Thanks for taking care of this and also tidying some things up in the api-docs 🧹 🚀

PR #1948, opened 2026-05-05
RISDEV-10227 expose paragraph switcher in single norms as landmark

andreasphil:


frontend/src/utils/search/dataFields.ts
UrsKahmann, line 1: Good catch! 🚀

UrsKahmann (commented):
🚀

PR #1951, opened 2026-05-05
RISDEV-10226 search pagination accessibility + page title

andreasphil:
- Updates the accessible labels of search pagination
- Focuses the first result after switching pages
- Inserts the page index into the title
- Fixes focus styles for norm search results

PR #1953, opened 2026-05-05
RISDEV-10226 simplify document type name mapping

andreasphil:
Addresses PR review feedback from #1951


UrsKahmann (approved):
Very nice 🚀

PR #1955, opened 2026-05-06
Fix wrapping of long article headings in norms view

UrsKahmann:
RISDEV-9451

### DIFF
--- a/frontend/src/components/documents/norms/LegislationContent.vue
+++ b/frontend/src/components/documents/norms/LegislationContent.vue
@@ -187,7 +187,7 @@ Attributes from the Juris CALS format without a corresponding HTML attribute mig
 
 /* Einzelvorschrift :: Überschrift  */
 :deep(h2.einzelvorschrift) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/documents/norms/LegislationContent.vue
```diff
@@ -187,7 +187,7 @@ Attributes from the Juris CALS format without a corresponding HTML attribute mig
 
 /* Einzelvorschrift :: Überschrift  */
 :deep(h2.einzelvorschrift) {
-  @apply ris-heading3-bold my-24 inline-block break-after-avoid;
+  @apply ris-heading3-bold my-24 inline-block break-after-avoid not-print:wrap-break-word not-print:hyphens-auto;
```
andreasphil, line 190: todo: why `not-print`? Should wrapping and hyphenation also work on print? Either way I would like us to adopt a "screen first" approach (similar to mobile first), so in this case something like: `wrap-break-word print:whatever-it-shouldbe`. It's more verbose but easier to understand in my opinion.

PR #1962, opened 2026-05-07
RISDEV-11567 page titles static content pages

UrsKahmann:
- this is a first part of page title adjustments focusing on the static pages
    - update the titles for the static pages according to the requirements
    - ogTitles stay untouched 
- refactors the `useStaticPageSeo` composable although I will probably merge this with the `useDynamicPageSeo` further down the line as they are very similar
- make use of nuxts `useSeoMeta` as that is already typed and easier to use than doing everything with `useHead`

## Question to the reviewer
1. I was wondering if we need tests for the static page titles but decided against it as adding e2e tests just for that seemed overkill. Wdyt?


frontend/src/composables/useStaticPageSeo.ts
```diff
@@ -1,24 +1,22 @@
-import { staticPageSeo } from "~/utils/i18n/staticPageSeo";
-import type { StaticPage } from "~/utils/i18n/staticPageSeo";
-
-export function useStaticPageSeo(page: StaticPage) {
-  const entry = staticPageSeo[page];
-
-  const title = entry.title;
-  const description = entry.description;
+export function useStaticPageSeo(
+  title: string,
```
andreasphil, line 2: todo: use an object with `title`, `description`, and `ogTitle` properties instead of 3 parameters, it's easier to read and extend.

andreasphil:
@UrsKahmann nice refactor with the `useSeoMeta` composable!

> I was wondering if we need tests for the static page titles but decided against it as adding e2e tests just for that seemed overkill. Wdyt?

Agree, but since you're also going to work on the page titles for dynamic pages (I think?): There it absolutely makes sense to test them, but some of them are also very complicated and add a lot of code to views. I would like to see them extracted into composables and have those composables unit tested. E2E tests are not the right place to test them (though I think we're doing this in some places currently).

frontend/src/composables/useStaticPageSeo.ts
UrsKahmann: Improved in https://github.com/digitalservicebund/ris-search/pull/1985

PR #1979, opened 2026-05-11
RISDEV-10430 Use playwright image in e2e test pipeline

UrsKahmann:
- run the e2e test job inside a playwright container, this way browser are already preinstalled
    - while this does not bring a significant speed up compared to specific runs, it should make the pipeline run time overall more consistent and prevent those super long running hiccups caused by the install browsers step 
- to prevent a docker in docker setup, opensearch is started as container service in parallel to the container job
- ~~update assertion so SEO test are run for the public prototype configuration instead of the private one~~
    - this breaks the tests - I think there are even more issues/questions regarding the SEO tests, will leave it like it is for now and bring up the SEO tests strategy as a separate topic 

RISDEV-10430


andreasphil (commented):
👍

PR #1982, opened 2026-05-11
RISDEV-11534 implement required skip links on all pages

andreasphil:


frontend/src/components/app/AppSkipLinks.vue
```diff
@@ -0,0 +1,21 @@
...
+  <nav
+    v-if="links?.length"
+    aria-label="Sprunglinks"
+    class="pointer-events-none fixed inset-x-0 top-8 z-10 flex justify-center gap-8 px-8"
+  >
+    <SkipLink
+      v-for="link in links"
+      :key="link.label"
+      :to="link.to"
+      class="pointer-events-auto"
```
UrsKahmann, line 16: **question:** why is this needed?

frontend/src/composables/useSimpleSearchRouteParams.ts
```diff
@@ -96,20 +97,35 @@ export function useSimpleSearchRouteParams() {
...
+      dateFilterTo: dateFilter.value.to ?? "",
+      pageIndex: pageIndex.value.toString(),
+      sort: sort.value,
+      itemsPerPage: itemsPerPage.value,
+    };
+
+    // Undefined values in the "to" query will be removed by the router
+    // automatically. Filtering them out here because we don't care about them
+    // but they might still cause comparison mismatches.
+    to = Object.fromEntries(
```
UrsKahmann, line 119: **question:** Why is this not needed in the `useAdavancedSearchRouteParams` ?

frontend/src/composables/useSkipLinks.ts
```diff
@@ -0,0 +1,72 @@
...
+  links: Readonly<Ref<SkipLink[]>>;
+
+  /**
+   * Adds a list of skip links to the page. Returns a cleanup method to remove
+   * them again.
+   */
+  register: (links: SkipLink[]) => () => void;
+};
+
+export const injectSkipLinksRegistry =
```
UrsKahmann, line 20: **suggestion:** Rename to something like `skipLinksRegistryInjectionKey` to better reflect that it is a key and not e.g. a function. `injectSkipLinksRegistry` sounds more like a function or method in my opinion.

frontend/src/composables/useSkipLinks.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!linksVal) return;
+
+    cleanup = registry.register(linksVal);
+  });
+
+  onUnmounted(() => {
+    cleanup?.();
+  });
+
+  return registry.links;
```
UrsKahmann, line 71: **question:** As far as I can see, the return value is not used anywhere, maybe it can be removed?

UrsKahmann (commented):
Really nice work 🚀 I like how you used the provider inject pattern - was able to learn a lot from this PR 🫶 Only left a few comments with mainly questions :)

frontend/src/composables/useSimpleSearchRouteParams.ts
andreasphil: because the advanced search currently doesn't have any parameters that can be `undefined`. If it has at some point we'll need to add it there too, but I didn't want to needlessly complicate it for now. Not sure if that's the best approach though 🤔

frontend/src/components/app/AppSkipLinks.vue
andreasphil: disabling pointer events on the parent means that we can still interact with content behind the container (skip links only occupy a small part in the middle of the container, but it's more convenient for layout to have the container stretch across the full page). So we need to enable pointer events again on the children to make them interactive

frontend/src/composables/useSkipLinks.ts
andreasphil: ah, that was intended for `AppSkipLinks` so it doesn't need to manually inject the registry, but I forgot to change it there. Will fix it

PR #1985, opened 2026-05-12
RISDEV 11568 simple search page titles

UrsKahmann:
### This PR:
- introduces a `useSeo` composable that will be used in all cases where static/dynamic page titles/meta information have to be set
- replaces the `useStaticPageSeo` composable with the `useSeo`, the `useDynamicPageSeo` will be replaced step by step in follow up PRs
- extracts the title composition logic for the simple search page title into a dedicated composable


frontend/src/composables/useSeo.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export function useSeo({ title, description, ogTitle }: SeoInput) {
+  const url = useRequestURL();
+
+  const resolvedTitle = computed(() => toValue(title));
+  const resolvedDescription = computed(() => toValue(description));
+  const resolvedOgTitle = computed(() => toValue(ogTitle ?? title));
+
+  useSeoMeta({
```
andreasphil, line 17: question: why the `resolved...` wrappers? Doesn't the `useSeoMeta` accept reactive properties?

UrsKahmann: Ahh yes, thats a leftover from a refactor, in my first implementation there was more logic (if empty -> undefined) in there, after removing that I overlooked that this doesn't make sense anymore 😅 thankssss, fill clean that up!

UrsKahmann: Fixed in https://github.com/digitalservicebund/ris-search/pull/1999

PR #1987, opened 2026-05-12
RISDEV-0000 put api docs updates in separate Dependabot group

andreasphil:


UrsKahmann (approved):
Looks good! nice idea to add the label 🚀

PR #1997, opened 2026-05-13
RISDEV-8808 redesigned back to top link

andreasphil:
reworked the back-to-top-link:

- in browsers with support for scroll animation timelines (currently: Chrome + Safari + FF Nightly), the button slides in over the first ~150px of scrolling
- otherwise the button is always shown
- also adds a `data-` attribute based approach for components to "nudge" the button a bit when its usual spot is occupied (e.g. bottom drawer). I chose this over a JavaScript solution to keep it simple and not cause issues with server-/client rendering, though admittedly it's a bit less explicit and relies on a small amount of global CSS

### DIFF
--- a/frontend/src/components/app/BackToTopLink.spec.ts
+++ b/frontend/src/components/app/BackToTopLink.spec.ts
@@ -9,7 +9,7 @@ describe("BackToTopLink", () => {
     await renderSuspended(BackToTopLink);
 
     expect(
-      screen.getByRole("link", { name: "zum Seitenanfang" }),
+      screen.getByRole("link", { name: "Zum Seitenanfang" }),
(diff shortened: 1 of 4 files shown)


UrsKahmann (approved):
Looks good! When the cookie banner is present its not scrolled to the very top, but I guess that is fine/expected as the banner doesn't count as main content?

andreasphil:
> Looks good! When the cookie banner is present its not scrolled to the very top, but I guess that is fine/expected as the banner doesn't count as main content?

I would say yes for now, as this is also the current behavior and no other behavior has been defined by the spec. I guess most people also don't leave the cookie banner hanging around for long so I consider that an edge cases. Good catch though, I missed that!

PR #2005, opened 2026-05-18
RISDEV-11642-dont-error-log-lucene-parser-errors

ekl176:
This PR stops logging lucene parser messages as errors by default. Instead of logging and returning a static error message we return the actual message of the parser exception.


UrsKahmann (approved):
Thanks for fixing this! 🚀

PR #2007, opened 2026-05-18
RISDEV-9453 fix skip link focus after navigation events

andreasphil:
- skip links no longer break on search pages when search query is completely empty (usually happens right after initial page load)
- ensures skip links are focused first after client-side navigation


UrsKahmann (commented):
👍

PR #2008, opened 2026-05-18
RISDEV-11570 norm page title adjustments

UrsKahmann:
## This PR:
- adapts title creation according to requirements
    - one part is still missing: using the ris:abkuerzung as fallback for the abbreviation, this will be done in a separate step as it involves backend changes as well.
- moves the title, ogTitle and description testing into the composable's unit tests and remove the e2e tests

NOTE: In the `buildOgTitle` I simplified the title composition. The way the `privateFeaturesEnabled` flag was used created a unreachable `else` block because in the prototype environment the validity period is always undefined (the documents don't contain them as they are considered private until going public). Thus its sufficient to just use the validity data if available (private features enabled) and omit them if not (e.g. in the prototype). Omitting none existing values in stead of the fallback strings (from the previous implementation that where never shown) was approved by the designers.

RISDEV-11570


andreasphil (approved):
😵‍💫

PR #2013, opened 2026-05-19
RISDEV 11570 caselaw page title

UrsKahmann:
- adapt construction of caselaw page title
- refactor meta tag creation and add tests
- the logic for the og description and title is rather hacky, I left it like it was for now as I didn't want to widen the scope of the ticket to much

RISDEV-11570


andreasphil (commented):
👍

PR #2014, opened 2026-05-19
RISDEV-0000 more CVE maintenance

andreasphil:


andreasphil:
causes problems and probably not worth it for the types of CVEs (nothing actually running in prod)

PR #2015, opened 2026-05-19
RISDEV 11570 single norm page title

UrsKahmann:
- adapt single norm page title to requirements
- refactor meta tag creation be reusing `useSeo` component
- extract common logic into a utility function
- add tests and remove e2e test (redundant)

RISDEV-11570


andreasphil (commented):
👍

PR #2018, opened 2026-05-20
Use playwright container for smoke tests

UrsKahmann:
- to stay consistent with the pipeline e2e test setup and reduce playwright browser download issues, use the playwright image for running smoke test as well
- [test run here ](https://github.com/digitalservicebund/ris-search/actions/runs/26161088737)succeeded and doesn't seem to take longer than before

RISDEV-0000


andreasphil (commented):
👍

PR #2023, opened 2026-05-22
RISDEV-0000 CVE maintenance

andreasphil:
- manually bumps the frontend Docker image version and some dependencies to get some upstream fixes for CVEs
- removes manual pinning and ignores that are no longer needed
- fixes some linting errors that started showing up after version upgrades


UrsKahmann (commented):
👍

PR #2041, opened 2026-05-27
RISDEV-11699 store search state directly in URL, remove sync

andreasphil:
This PR refactors both the simple and the advanced search to store their state directly in the URL, removing the intermediary state and the syncing of that state with the URL. This addresses various issues where the browser navigation functionality was broken because of race conditions and other unhandled edge cases in the synchronization logic.

For this it:

- Replaces the logic for reading the URL state with computed properties, returning the state directly from the URL
- Removes the watchers (no longer needed thanks to the computer properties)
- Updates the logic for saving state to the URL
- Updates the simple and advanced search to use the new properties
  - In some cases where we don't want the state of some user interaction to be synced into the URL immediately, The view now contains a temporary in-between state that is saved to the URL once the interaction completes.

### DIFF
--- a/frontend/e2e/advancedSearch.spec.ts
+++ b/frontend/e2e/advancedSearch.spec.ts
@@ -89,6 +89,29 @@ test.describe("general advanced search page features", () => {
     ).toBeVisible();
   });
 
+  test("browser back restores previous state", async ({ page }) => {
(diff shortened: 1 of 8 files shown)


UrsKahmann (approved):
Great work! 🚀 👏

PR #2042, opened 2026-05-27
RISDEV-11644-remove-locking-mechanism

ekl176:
This PR removes the locking mechanism from the indexing state file. 
Locking the index manually became obsolete after switching to Kubernetes jobs to trigger the index.


UrsKahmann (approved):
Thanks for cleaning this up 🚀 
Some small things left to do:
- **nitpick**: Many of the comments in the `IndexSyncJob.java` and one in `IndexingState.java` still mention the locking mechanism
- **todo:** `PersistedIndexingState.java` (not sure if or where that is used) still contains a `lockTime` parameter.. probably not needed anymore? Maybe check if the whole class can be removed, looks like its not used anywhere 🤔

PR #2043, opened 2026-05-28
RISDEV-11501 cleanup api docs

UrsKahmann:
**simplifies the api docs application by:**
- using a plain vitepress project as base
- try to use as little customization as possible 
    - added button styling (based on the vitepress button style)
    - footer and feedback form
- removed pages that don't add value 
- used as little dependencies as possible
- added linting and formatting (also checked in the pipeline)
- building doesn't produce any errors anymore

RISDEV-11501


api-docs/.vitepress/theme/components/SwaggerButton.vue
```diff
@@ -1,15 +1,10 @@
...
-import { useData } from "vitepress";
-import ButtonLink from "./ButtonLink.vue";
-const { theme } = useData();
-</script>
-
 <template>
-  <ButtonLink
-    :href="theme.swaggerUiUrl"
+  <a
+    class="RisButton"
```
andreasphil: nitpick: `ris-button` would be the more idiomatically CSS naming for this class

api-docs/.vitepress/theme/components/FeedbackForm.vue
```diff
@@ -1,82 +1,108 @@
...
+  "There has been an error submitting your feedback. Please try again later.";
 
 const submitFeedback = async () => {
-  errorMessage.value = undefined
-  if (isStringEmpty(feedback.value)) {
-    errorMessage.value = emptyMessageError
-    return
+  errorMessage.value = undefined;
+  if (!feedback.value.trim()) {
+    errorMessage.value = "Please enter your feedback in the box provided";
```
andreasphil: nitpick: the phrasing sounds a bit strange to me, maybe "Please enter your feedback in the provided box" or simply "Please enter your feedback"? (I know it was like this before)

api-docs/.vitepress/theme/components/FeedbackForm.vue
```diff
@@ -1,82 +1,108 @@
...
-  isSent.value = true;
-}
+};
 </script>
 
 <template>
-  <div class="flex max-w-prose flex-col">
-    <div v-if="isSent" class="space-y-24" data-test-id="feedback-sent-confirmation">
-      <div class="ris-heading3-bold">Thank you for your Feedback!</div>
+  <div class="feedback-form">
```
andreasphil: todo:

- use [CSS modules](https://vuejs.org/api/sfc-css-features.html#css-modules)
- don't use BEM - this is a legacy naming convention from before we had proper scoping in CSS, I wouldn't want to introduce that into a modern code base

api-docs/.vitepress/theme/components/FeedbackForm.vue
```diff
@@ -1,82 +1,108 @@
...
+      <p>
+        Your feedback will help us to improve the documentation. Describe your
+        request in as much detail as possible. Do not enter any personal data.
+      </p>
+      <div class="feedback-form__field">
+        <textarea
+          v-model="feedback"
+          placeholder="Enter feedback"
+          rows="6"
+          :class="{ 'feedback-form__textarea--invalid': errorMessage }"
```
andreasphil: suggestion: for validation, look into

- [Constraint Validation API](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Constraint_validation)
- [`:user-invalid`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:user-invalid)

I think the type of validation that we need here can be achieved without JavaScript.

api-docs/.vitepress/theme/components/Footer.vue
```diff
@@ -0,0 +1,80 @@
...
+          <a href="https://digitalservice.bund.de" target="_blank"
+            >DigitalService GmbH des Bundes</a
+          >
+        </p>
+      </div>
+    </div>
+  </footer>
+</template>
+
+<style scoped>
```
andreasphil: todo: here too, use CSS modules

api-docs/.vitepress/theme/custom.css
```diff
@@ -0,0 +1,49 @@
...
+  display: inline-block;
+  text-align: center;
+  font-weight: 600;
+  white-space: nowrap;
+  text-decoration: none;
+  transition:
+    color 0.25s,
+    border-color 0.25s,
+    background-color 0.25s;
+  padding: 0 20px;
```
andreasphil: todo: use rem equivalents so it scales properly when users override the font size

api-docs/.vitepress/theme/Layout.vue
```diff
@@ -0,0 +1,46 @@
...
+        <FeedbackForm />
+      </div>
+      <div class="vp-doc footer">
+        <Footer />
+      </div>
+    </template>
+  </Layout>
+</template>
+
+<style scoped>
```
andreasphil: todo: use CSS modules

api-docs/.oxlintrc.json
```diff
@@ -0,0 +1,28 @@
...
+    "typescript",
+    "unicorn",
+    "vue"
+  ],
+  "categories": {
+    "correctness": "error",
+    "suspicious": "error"
+  },
+  "rules": {
+    "import/no-unassigned-import": "off",
```
andreasphil, line 26: todo: can we also have the comments here for why the change is necessary? I know it's pretty obvious in these cases, but I would like to set the bar for deviating from the standard a little bit higher than simply adding a rule without explanation :D

andreasphil (changes requested):
Thank you so much @UrsKahmann, looking really good! A few comments and required changes but only smaller things

api-docs/.vitepress/theme/components/FeedbackForm.vue
UrsKahmann: Or would it maybe even be better to disable the button as long as the textfield is empty or contains only whitespace? 

EDIT: Never mind,  thats not such a great user experience I guess 🙃

andreasphil (approved):

PR #2044, opened 2026-05-29
RISDEV-10551 Align items per page options across searches

UrsKahmann:
- align items per page options across searches
- set options to 25, 50 and 100

RISDEV-10551


andreasphil (commented):
👍

PR #2052, opened 2026-06-02
RISDEV-8269 date filter fassungen

UrsKahmann:
## This PR:
- adds a date filter for the fassungen table
- the clear button is not yet implemented, will be done in a separate PR

## Notes
- I really struggled with the date comparison and parsing
    - parseDateGemanLocalTime fails if the year is e.g. 0300 (not a real use case but still shouldn't throw)
    - I think the date handling in general needs a refactor as we only want to work with dates (timeless) but using dayjs makes this quite hard. We should probably consider switching to the new temporal API

### DIFF
--- a/frontend/src/components/documents/norms/NormVersionList.vue
+++ b/frontend/src/components/documents/norms/NormVersionList.vue
@@ -90,7 +90,13 @@ async function handleSelectionUpdate(newSelection: TableRowData) {
     :row-class="rowClass"
     @row-select="onRowSelect"
     @update:selection="handleSelectionUpdate"
+    :pt="{
+      emptyMessageCell: {
+        class: 'ps-16 py-12 text-left text-gray-900',
+      },
+    }"
   >
+    <template #empty>Keine Ergebnisse gefunden</template>
     <Column
       field="fromDate"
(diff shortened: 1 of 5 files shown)


frontend/src/components/documents/norms/NormVersionList.vue
```diff
@@ -90,7 +90,13 @@ async function handleSelectionUpdate(newSelection: TableRowData) {
     :row-class="rowClass"
     @row-select="onRowSelect"
     @update:selection="handleSelectionUpdate"
+    :pt="{
```
andreasphil, line 93: comment: Usually I would expect this type of thing to go into RIS UI. However, since we're going to make a lot of changes to the versions list anyway, we don't need to do that here

andreasphil (commented):
👍

frontend/src/components/documents/norms/NormVersionList.vue
UrsKahmann: Thanks you are right. I think Zarah also already pointed out that the text color for the empty table hint should be black. So when I change that I could also move this to ris-ui..

PR #2053, opened 2026-06-03
RISDEV-11699 fix broken SSR for skip links

andreasphil:
This PR refactors the skip links to use Nuxt's built in `definePageMeta` for defining skip links per page, and removes the custom composable. This simplifies the code and fixes an issue where skip links would not be rendered on the server because of the order in which elements are rendered.


frontend/src/components/app/SkipLinks.spec.ts
```diff
@@ -1,37 +1,50 @@
...
+    await renderSuspended(AppSkipLinks, {
+      props: {
+        links: [{ label: "Extra", to: "#extra" }],
+      },
+      route: "/no-links-here",
+    });
+
+    // Only prop links visible (no route meta on this route)
+    expect(screen.getByRole("link", { name: "Extra" })).toBeVisible();
+  });
```
UrsKahmann, line 32: **question:** The test title doesn't seem to match what is actually tested (or I am not understanding it 🙃 ). I would have expected the route meta to also carry a skiplink

UrsKahmann (commented):
Niceeee 🚀

frontend/src/components/app/SkipLinks.spec.ts
andreasphil: whooops yes seems like I mixed something up with a previous iteration of the component interface, I'll fix it!

PR #2061, opened 2026-06-04
RISDEV-11384 styling for pre-formatted footnotes

andreasphil:


andreasphil:
@UrsKahmann decided to keep the font definition in this project instead of RIS UI since I'm not aware of a need for this anywhere else, and I don't trust the definitions to be "stable" yet 🙊

UrsKahmann (commented):
Looking good 👍

PR #2065, opened 2026-06-05
RISDEV-11758-handle-client-abort-exception

ekl176:
We don't want to log errors when a client closes the connection. We also don't need to return any response, since the connection is lost already.

The ClientAbortException will be wrapped in a HttpMessageNotWritableException, so we need to parse this one in the ControllerAdvice.


andrew-cenkner-digitalservice (commented):
Approved

PR #2076, opened 2026-06-08
Risdev 11293 only consider roots as being deleted

ekl176:
The Changelogs show all changes down to a specific file. For the changelog api we want to narrow changes down to the identifier. 

We will only add files to the deleted list of an api changelog when it is the file that we consider the root document.

### DIFF
--- a/backend/src/test/java/de/bund/digitalservice/ris/search/unit/mapper/ChangelogResponseMapperTest.java
+++ b/backend/src/test/java/de/bund/digitalservice/ris/search/unit/mapper/ChangelogResponseMapperTest.java
@@ -1,0 +1,65 @@
+package de.bund.digitalservice.ris.search.unit.mapper;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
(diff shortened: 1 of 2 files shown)


backend/src/test/java/de/bund/digitalservice/ris/search/unit/mapper/ChangelogResponseMapperTest.java
```diff
@@ -0,0 +1,65 @@
...
+import de.bund.digitalservice.ris.search.models.DocumentKind;
+import de.bund.digitalservice.ris.search.schema.ChangelogResponse;
+import java.util.HashSet;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+class ChangelogResponseMapperTest {
+
```
UrsKahmann, line 17: **todo:** Add tests that test that when the root docuement is deleted it is correctly reflected as deleted in the repsonse.

UrsKahmann (approved):

PR #2086, opened 2026-06-10
RISDEV-8269 Add clear button to DateInput

UrsKahmann:
## This PR:
- introduces a wrapper component to overlay a "clear" button on the date input field
- enables the clear button for the fassungen date filter in the fassungen tab

## Questions to the reviewer:
I am not super happy with the implementation, it feels unclean/hacky, especially:
- how the button is positioned and sized -> Is it bad to give it a fixed size in em? 
    - I would have liked to place it vertically centered in on the component but that didn't really work because in case of an error the date input shows an error message and that needs to be the direct sibling of the date input to be correclty styled ... thus the component grows when the error message is shown -> centering vertically offsets the button from the actual input field .. thats why i ended up pinning it to the top right corner
- I am not sure how "reusable" the wrapper actually is in its current state .. does it even make sense to have it in a separate component as of now?

All feedback and suggestions for improvement are welcome 🙏 

RISDEV-8269


frontend/src/components/DateInput.vue
```diff
@@ -36,13 +37,17 @@ const props = withDefaults(
 
     /** Validation error and message to display. */
     validationError?: ValidationError;
+
+    /** Whether to show a clear button. */
```
andreasphil, line 41: suggestion: `showClear` (same naming as in PrimeVue components that support this)

frontend/src/components/DateInput.vue
```diff
@@ -167,13 +174,27 @@ function focus() {
...
+  await nextTick();
+  focus();
+}
+
 defineExpose({ focus });
 </script>
 
 <template>
-  <div class="flex w-full flex-col gap-2">
+  <ClearButtonWrapper
```
andreasphil: suggestion: move the error message outside the slot and render it next to the `ClearButtonWrapper`. That way the `ClearButtonWrapper` doesn't need to be concerned with layout shifts/resizing. Some other changes that this would enable:

- in the clear button component, center the button in the container instead of positioning it absolutely
- instead of giving the button a fixed size, calculate the height (like `100% - 8px`) + aspect ratio of 1 to get the width right

You can decide which one you like better, but worth looking into.

**Edit:** I missed your comment that the styling of the error message depends on it being a sibling of the input. We should still weigh the options and but then also consider that we would need to manually add the styling for the error message. Since it's not a lot that might be acceptable?

frontend/src/components/DateInput.vue
```diff
@@ -167,13 +174,27 @@ function focus() {
...
+    :clearButtonVisible="!!inputValue && props.withClearButton"
+    @clear="clear"
+  >
     <InputMask
```
andreasphil: todo: this needs a padding on the right side with the size of the clear button, otherwise the button might obscure part of the input depending on the size. Not an issue that we currently have with how we're using the component, but also not something that we should rely on.

frontend/src/components/WithClearButton.vue
```diff
@@ -0,0 +1,36 @@
+<script setup lang="ts">
```
andreasphil, line 1: suggestion: name this component `WithClearButton`

```html
<WithClearButton>
  <InputMask />
</WithClearButton>
```

... reads pretty nice I think :)

frontend/src/components/DateInput.vue
UrsKahmann: Ahh yes good catch! But I am unsure about the best way to apply is.. I think ideally the wrapper would somehow add the padding as it already knows its size and whens its shown or not? But I have now idea how to correctly pass that padding to the wrapped element 🤔  
Or I simply apply the padding here directly to the DateInput but only if the clear button is shown? And how would I get the size right? 
Would love to hear your thoughts @andreasphil  🙏

UrsKahmann: I went with the simplest approach: https://github.com/digitalservicebund/ris-search/pull/2086/commits/a631ab73defefba7d24468f13557b01e38759ad4 let me know if thats baaad : D

PR #2092, opened 2026-06-12
RISDEV-11775 prototype on-page text search

andreasphil:
This PR demonstrates what a basic implementation of a <kbd>ctrl + f</kbd>-style text search on the page could look like. Don't merge pls 🙇‍♂️

Based on the search example in the [Custom CSS Highlights API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API#:~:text=white%3B%0A%7D-,Result,-The%20result%20is) docs on MDN.

### DIFF
--- a/frontend/src/assets/main.css
+++ b/frontend/src/assets/main.css
@@ -42,6 +42,11 @@ mark {
   @apply bg-yellow-200;
 }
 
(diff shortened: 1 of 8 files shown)


andreasphil:

PR #2110, opened 2026-06-15
RISDEV-0000 Fix frontend CVEs in sub-dependencies 

UrsKahmann:
- also fix node version in test pipelines

RISDEV-0000


andreasphil (commented):
👍

PR #2112, opened 2026-06-15
RISDEV-0000 Try using dependabot for image updates

UrsKahmann:
- it seems like dependabot is now able to update chainguard images 
- this PR disables the custom update action and enables dependabot image updates for all applications - lets see if that works

RISDEV-0000

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -60,6 +60,3 @@ updates:
       day: "sunday"
     cooldown:
       default-days: 3
-    ignore:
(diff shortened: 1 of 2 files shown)


andreasphil (commented):
Change LGTM but no idea if this is going to work, let's see 😅 should be easy enough to revert if it doesn't

UrsKahmann:
> Change LGTM but no idea if this is going to work, let's see 😅 should be easy enough to revert if it doesn't

Seems to be working: https://github.com/digitalservicebund/ris-search/pull/2118 🎉

PR #2114, opened 2026-06-15
RISDEV-10395 clean up and simplify search result teasers

andreasphil:
In preparation of RISDEV-10395, this PR tidies the search result teasers:

- extracted shared search result logic for matches and headings into a new module `utils/search/searchResults.ts`
- added a shorthand `stripAllHtml` sanitizer for when all HTML should be removed
- all props now use prop destructuring for more concise code
- removed unnecessary `props.` prefix from template code
- moved computed properties into a consistent, more logical order
- moved all sanitization into the setup script instead of mixing it between templates and script
- removed intermediary `metadata` computed objects, flattening them to simple computed properties
- switched to autoimports where possible
- reduced API surface by merging all reactive properties that were only referenced by other reactive properties
- removed `...as SomeType` casts in favor of typed mapping in some places; fixes some issues where the cast would mask that a value may be `undefined`

### DIFF
--- a/frontend/src/components/search/NormSearchResult.vue
+++ b/frontend/src/components/search/NormSearchResult.vue
@@ -1,81 +1,70 @@
 <script setup lang="ts">
+import type { Dayjs } from "dayjs";
 import IcBaselineBalance from "~icons/ic/baseline-balance";
 import type { SearchResultHeaderItem } from "~/components/search/SearchResultHeader.vue";
-import type {
-  LegislationExpression,
-  SearchResult,
-  TextMatch,
-} from "~/types/api";
+import type { LegislationExpression, SearchResult } from "~/types/api";
+import { getMatch, getTitleWithFallback } from "~/utils/search/searchResults";
 
-const props = defineProps<{
+const { searchResult, order } = defineProps<{
   searchResult: SearchResult<LegislationExpression>;
   order: number;
 }>();
 
 const { searchResultClicked } = usePostHog();
 
-const item = computed(() => props.searchResult.item);
-
-const textMatches: ComputedRef<TextMatch[]> = computed(
-  () => props.searchResult.textMatches,
-);
-
-const headline = computed(() => {
-  const match =
-    getMatch("name", textMatches.value) ||
-    item.value.name ||
-    "Titelzeile nicht vorhanden";
-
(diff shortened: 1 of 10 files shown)


frontend/src/components/search/NormSearchResult.vue
```diff
@@ -1,89 +1,70 @@
...
-    "Titelzeile nicht vorhanden";
+const resultTypeId = useId();
+
+const headerItems = computed<SearchResultHeaderItem[]>(() => {
+  let date: string | Dayjs | undefined =
+    searchResult.item?.exampleOfWork.legislationDate;
+
+  if (privateFeaturesEnabled) {
+    const coverage = temporalCoverageToValidityInterval(
+      searchResult.item?.temporalCoverage,
```
UrsKahmann, line 32: **nitpick:** I think the `?` is not needed here.

frontend/src/components/search/NormSearchResult.vue
```diff
@@ -1,89 +1,70 @@
...
-const headline = computed(() => {
-  const match =
-    getMatch("name", textMatches.value) ||
-    item.value.name ||
-    "Titelzeile nicht vorhanden";
+const resultTypeId = useId();
+
+const headerItems = computed<SearchResultHeaderItem[]>(() => {
+  let date: string | Dayjs | undefined =
+    searchResult.item?.exampleOfWork.legislationDate;
```
UrsKahmann, line 28: **nitpick:** I think the `?` is not needed here.

frontend/src/utils/search/searchResults.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+        "AZ <mark>456</mark>",
+      ]);
+    });
+
+    it("returns only the sanitized texts for the requested name", () => {
+      expect(getMatches("headline", matches)).toEqual([
+        "The <mark>Main</mark> Title",
+        "Second headline",
+      ]);
+    });
```
UrsKahmann, line 53: **suggestion:** Remove one of the tests, they seem redundant?

UrsKahmann (commented):
Nice refactor and cleanup 🚀  I like that you move common logic into the searchResults module 🙌

andreasphil:
Review comments addressed in #2127

PR #2115, opened 2026-06-15
Risdev 11525 use xslt library for caselaw

MagdaN:
This PR:
- removes the local version of the xslt and uses the one from the sahred library
- adapts e2e testsfiles
- adapts all backend builds in workflows in order to use a github token to fetch the library from the github maven registry

Pease be aware:
- this should only be merged when the caselaw data in the bucket has been updated
- after this has been merged you need a github token in your local env in order to build the backend 🫣


backend/Dockerfile.prod
```diff
@@ -9,4 +16,6 @@ ARG SENTRY_RELEASE
 COPY --from=backend-builder /app/build/libs/ris-search-*.jar /app/app.jar
 EXPOSE 8090
 ENV SENTRY_RELEASE=${SENTRY_RELEASE}
+ENV GH_PACKAGES_REPOSITORY_USER=${GH_PACKAGES_REPOSITORY_USER}
```
UrsKahmann: **todo:** Remove the ENV vars here, they shouldn't be needed as the build is already done in the first stage?

backend/Dockerfile.prod
```diff
@@ -9,4 +16,6 @@ ARG SENTRY_RELEASE
...
+ENV GH_PACKAGES_REPOSITORY_TOKEN=${GH_PACKAGES_REPOSITORY_TOKEN}
```
UrsKahmann: **todo:** Same here, remove this.

backend/Dockerfile.prod
```diff
@@ -1,6 +1,13 @@
 FROM gradle:9.5.1-jdk21@sha256:4a011ed0edfe2feb7dace27d1fedf22caf217918b59932a95605cb7beec142c4 AS backend-builder
 WORKDIR /app
 COPY . /app
+
+ARG GH_PACKAGES_REPOSITORY_USER
+ARG GH_PACKAGES_REPOSITORY_TOKEN
+
+ENV GH_PACKAGES_REPOSITORY_USER=${GH_PACKAGES_REPOSITORY_USER}
+ENV GH_PACKAGES_REPOSITORY_TOKEN=${GH_PACKAGES_REPOSITORY_TOKEN}
```
UrsKahmann: **todo:** Instead of using ARG and ENV, [mount the secret into the build stage](https://docs.docker.com/build/building/secrets/#using-build-secrets). See how its done in the frontend for reference.

backend/Dockerfile
```diff
@@ -1,5 +1,12 @@
 FROM gradle:9.5.1-jdk21@sha256:4a011ed0edfe2feb7dace27d1fedf22caf217918b59932a95605cb7beec142c4
 WORKDIR /backend
+
+ARG GH_PACKAGES_REPOSITORY_USER
+ARG GH_PACKAGES_REPOSITORY_TOKEN
+
+ENV GH_PACKAGES_REPOSITORY_USER=${GH_PACKAGES_REPOSITORY_USER}
+ENV GH_PACKAGES_REPOSITORY_TOKEN=${GH_PACKAGES_REPOSITORY_TOKEN}
```
UrsKahmann, line 8: **todo:** Same here, although this is not intended for the production build, its better to [use secret mounts](https://docs.docker.com/build/building/secrets/#using-build-secrets) for passing secrets into the build.

backend/build.gradle.kts
```diff
@@ -96,6 +103,10 @@ dependencies {
     // CVE-2026-41293, CVE-2026-43512, CVE-2026-41284, CVE-2026-42498, CVE-2026-43513
     implementation(libs.tomcat.embed)
 
+    implementation(libs.ris.html.transformation)
+    // Local development:
+    // implementation(files("../../ris-html-transformation/build/libs/ris-html-transformation-0.0.1-SNAPSHOT.jar"))
```
UrsKahmann: **todo:** Remove this comment as its not needed anymore right?

backend/build.gradle.kts
```diff
@@ -31,6 +31,13 @@ configurations {
 
 repositories {
     mavenCentral()
+    maven {
+        url = uri("https://maven.pkg.github.com/digitalservicebund/ris-html-transformation")
+        credentials {
+            username = System.getenv("GH_PACKAGES_REPOSITORY_USER")
+            password = System.getenv("GH_PACKAGES_REPOSITORY_TOKEN")
+        }
```
UrsKahmann, line 39: **question:** Does this mean to build the backend locally we will always need to have these env vars set with a valid token? Or is this only needed for the e2e test pipeline?

UrsKahmann (changes requested):
Looking good already, I only focused on the pipeline and docker file changes with my review :)

backend/build.gradle.kts
MagdaN: gives you a hint how to develop locally? But I can remove it

backend/build.gradle.kts
MagdaN: yes unfortunately that is needed, if you don't use your local version

PR #2124, opened 2026-06-16
RISDEV-8269 Simplify DateInput component error logic

UrsKahmann:
- remove validation error model as it was not used anywhere to set the error from outside the component and made the whole logic more complex
- this resolves a bug where typing after entering an invalid date changed the error message from "Kein valides Datum" to "Ungültige Eingabe"

RISDEV-8269

### DIFF
--- a/frontend/src/components/DateInput.vue
+++ b/frontend/src/components/DateInput.vue
@@ -126,7 +126,7 @@ defineExpose({ focus });
         ref="inputMaskEl"
         v-model="inputValue"
         :auto-clear="false"
-        :invalid="errorMessage != undefined"
+        :invalid="errorMessage !== undefined"
         :readonly="isReadOnly"
         :disabled="isReadOnly"
         class="w-full"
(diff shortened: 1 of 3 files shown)


frontend/src/components/DateInput.vue
```diff
@@ -199,15 +126,14 @@ defineExpose({ focus });
         ref="inputMaskEl"
         v-model="inputValue"
         :auto-clear="false"
-        :invalid="effectiveHasError"
+        :invalid="errorMessage != undefined"
```
andreasphil: nitpick:

```suggestion
        :invalid="errorMessage !== undefined"
```

andreasphil (approved):
Great find and thanks for refactoring. Some context: We inherited this component from the other NeuRIS streams. The validation error type and its related logic were required for server side validation. This is not a use case that we currently have and probably not one that we will ever have (and even if we do, there's probably a better way of implementing it). So I agree this is unnecessary complexity and can be removed. Thank you!

PR #2125, opened 2026-06-16
refactor:risdev-11832 bulk save documents in base in dex service

ekl176:
This pr changes how one batch of documents is saved to opensearch. Instead of parsing and saving every document one by one, it parses and then stores them in batches of 100.


MagdaN (approved):

UrsKahmann (approved):
Looks good to me 🚀  I am excited to see how that speeds up things 🕐 🎉

PR #2128, opened 2026-06-17
refactor:risdev-11833 refactor norms reindex

ekl176:
The goal of this PR is to reduce the number of list calls to s3 during a full reindex. In its current implementation, we make 2 additional s3 calls per document work after listing all files. When keeping the files in memory we can reduce this overhead completely. 

The works are still processed sequentially right now but instead of storing every expression of a work one after the other, we store the expressions in bulk.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
@@ -17,7 +17,6 @@ import java.time.LocalDate;
 import java.time.Month;
 import java.time.Period;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -27,7 +26,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.apache.commons.collections4.ListUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -75,7 +73,8 @@ public class IndexNormsService implements IndexService {
    */
   public void reindexAll(String startingTimestamp) {
     DateUtils.avoidOpenSearchSubMillisecondDateBug();
-    Set<WorkEli> workElis = getWorks(normsBucket.getAllKeysByPrefix("eli/").stream());
+    List<String> allFiles = normsBucket.getAllKeysByPrefix("eli/");
+    Map<WorkEli, List<String>> workElis = getWorks(allFiles.stream());
     processWorkEliUpdates(workElis, startingTimestamp);
     clearOldNorms(startingTimestamp);
   }
@@ -83,7 +82,7 @@ public class IndexNormsService implements IndexService {
   @Override
   public void indexChangelog(Changelog changelog) {
     try {
-      Set<WorkEli> workElis =
+      Map<WorkEli, List<String>> workElis =
           getWorks(Stream.concat(changelog.getChanged().stream(), changelog.getDeleted().stream()));
       processWorkEliUpdates(workElis, Instant.now().toString());
     } catch (IllegalArgumentException e) {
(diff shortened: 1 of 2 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
```diff
@@ -101,21 +109,29 @@ private Set<WorkEli> getWorks(Stream<String> files) {
         .collect(Collectors.toSet());
   }
 
-  private void processWorkEliUpdates(Collection<WorkEli> workElis, String startingTimestamp) {
-    List<List<WorkEli>> batches = ListUtils.partition(workElis.stream().toList(), 100);
-    for (int i = 0; i < batches.size(); i++) {
-      List<WorkEli> oneBatch = batches.get(i);
-      logger.info("Indexing batch {} of {}", i + 1, batches.size());
```
andrew-cenkner-digitalservice, line 108: I think we lost the logging in this refactor. To me the logging is important because I look at these logs every time I do a reindex all.
Could you please add the logging back for the new version?

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
```diff
@@ -212,12 +226,12 @@ private static boolean norm2StartsBeforeNorm1Ends(Norm norm1, Norm norm2) {
     return !norm2.getEntryIntoForceDate().isAfter(norm1.getExpiryDate());
   }
 
-  private Optional<Norm> getNormFromS3(ExpressionEli expressionEli)
+  private Optional<Norm> getNormFromS3(ExpressionEli expressionEli, List<String> filenames)
```
andrew-cenkner-digitalservice, line 244: Since this doesn't actually call S3 anymore, we should change the name of the method.

backend/src/test/java/de/bund/digitalservice/ris/search/unit/service/IndexNormsServiceTest.java
```diff
@@ -133,23 +132,21 @@ void reindexAllIgnoresInvalidFiles() throws ObjectStoreServiceException {
...
-        .thenReturn(
-            List.of("eli/bund/bgbl-1/1992/s101/1992-01-01/1/deu/1992-01-02/regelungstext-1.xml"));
     when(this.bucket.getFileAsString(
             "eli/bund/bgbl-1/1992/s101/1992-01-01/1/deu/1992-01-02/regelungstext-1.xml"))
         .thenReturn(Optional.of(testContent));
 
     String startingTimestamp = "2024-01-01T12:00:00Z";
     this.service.reindexAll(startingTimestamp);
 
-    verify(repo, times(1)).save(any());
```
andrew-cenkner-digitalservice, line 150: Is it on purpose that we no longer check that save (now saveall) is called exactly once? The next verification only checks it was called that way, not that it was not called other ways.

andrew-cenkner-digitalservice (changes requested):

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
ekl176: it still gets a norm from s3, but we do not call a 'list objects' to retrieve manifestations beforehand.

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
```diff
@@ -85,7 +84,16 @@ public void indexChangelog(Changelog changelog) {
...
           getWorks(Stream.concat(changelog.getChanged().stream(), changelog.getDeleted().stream()));
-      processWorkEliUpdates(workElis, Instant.now().toString());
+
+      // retrieve all file paths for the given workElis
+      Map<WorkEli, List<String>> allFilesByWorkEli =
+          workElis.stream()
+              .collect(
+                  Collectors.toMap(
+                      workEli -> workEli,
+                      workEli -> normsBucket.getAllKeysByPrefix(workEli.toString() + "/")));
```
andrew-cenkner-digitalservice: I would prefer
```
        Map<WorkEli, List<String>> allFilesByWorkEli = 
                Maps.asMap(workElis, workEli -> normsBucket.getAllKeysByPrefix(workEli.toString() + "/"));
```
or
```
Map<WorkEli, List<String>> allFilesByWorkEli = new HashMap<>();
for (WorkEli workEli : workElis) {
    allFilesByWorkEli.put(workEli, normsBucket.getAllKeysByPrefix(workEli.toString() + "/"));
}
```
but that's just a style preference.

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
ekl176: Putting some logging back in is a good idea. I would try to log chunks without having to create the partitions though. We actually don't process anything in batches so we might get away with a more efficient way to log this.

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
ekl176: the iterative approach might indeed be easier to read. 

i think the Maps.asMap will behave differently as it evaluates lazily. Since we want to pass around the paths inside the indexing service, I'd rather create them up front to avoid calling getAllKeysByPrefix multiple times by accident.

backend/src/test/java/de/bund/digitalservice/ris/search/unit/service/IndexNormsServiceTest.java
ekl176: no on purpose. its back in

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
ekl176: replaced the stream with a for each to build the map.

backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexNormsService.java
```diff
@@ -101,21 +112,41 @@ private Set<WorkEli> getWorks(Stream<String> files) {
...
+      Map<WorkEli, List<String>> workElis, String startingTimestamp) {
+    int processedWorkEli = 0;
+    int totalWorkElis = workElis.size();
+
+    for (Map.Entry<WorkEli, List<String>> entry : workElis.entrySet()) {
+      processOneNormWork(entry.getKey(), entry.getValue(), startingTimestamp);
+
+      processedWorkEli++;
+      if (processedWorkEli % 100 == 0 || processedWorkEli == totalWorkElis) {
+        logger.info("index progress: {}/{} works processed", processedWorkEli, totalWorkElis);
```
andrew-cenkner-digitalservice, line 140: I would have a slight preference to change the method signature so the code would be
```
  private void processWorkEliUpdates(
      List<Map.Entry<WorkEli, List<String>>> entries, String startingTimestamp) {

    for (int i = 0; i < entries.size(); i++) {
      Map.Entry<WorkEli, List<String>> entry = entries.get(i);
      processOneNormWork(entry.getKey(), entry.getValue(), startingTimestamp);
      if (processedCount % 100 == 0) {
        logger.info("index progress: {}/{} works processed", processedCount, entries.size());
      }
    }
    logger.info("indexing complete.");
  }
```
but it's just a style preference, so I will approve the PR.

andrew-cenkner-digitalservice (approved):

PR #2129, opened 2026-06-17
RISDEV-10822 Frontend breakpoint alignments

UrsKahmann:
- align breakpoint definitions with design requirements
- replace usage of tailwind _container_ utility with simpler wrapper class

RISDEV-10822

### DIFF
--- a/frontend/src/assets/main.css
+++ b/frontend/src/assets/main.css
@@ -32,8 +32,10 @@
   }
 }
 
-@utility container {
-  @apply mx-auto px-16;
+@layer utilities {
+  .wrapper {
+    @apply mx-auto px-16 lg:px-24 2xl:max-w-[96rem];
+  }
 }
 
 mark {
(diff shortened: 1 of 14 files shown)


frontend/src/assets/main.css
```diff
@@ -24,8 +31,10 @@
   }
 }
 
-@utility container {
-  @apply mx-auto px-16;
+@layer utilities {
+  .wrapper {
+    @apply mx-auto px-16 lg:px-24 2xl:max-w-[96rem];
```
andreasphil: todo: [reference the breakpoint variable](https://tailwindcss.com/docs/max-width#using-a-custom-value)

```suggestion
    @apply mx-auto px-16 lg:px-24 2xl:max-w-(--breakpoint-2xl);
```

andreasphil (changes requested):
just a small (but required) change request 👍

PR #2131, opened 2026-06-18
fix: risdev-10395-only show article texthighlights

ekl176:


frontend/e2e/simpleSearch.spec.ts
```diff
@@ -273,35 +273,9 @@ test.describe("searching legislation", () => {
       }),
     ).toBeVisible();
 
-    // Highlights
     await expect(
```
andreasphil, line 322: todo: this test is supposed to test that highlighting and displaying sections with matches works, so the correct approach for making this test pass again would be to change the query to one that has highlights and then adjust the assertions. Otherwise that isn't covered by tests anymore

andreasphil (changes requested):

frontend/e2e/simpleSearch.spec.ts
ekl176: I'll also add a test then to reflect the new behaviour

PR #2132, opened 2026-06-19
RISDEV-10826 adjust responsive typography styles for content pages

andreasphil:


UrsKahmann (commented):
Looks good! 🙌 
I noticed that the H1 is a tiny bit misaligned to the right:

Image

Maybe thats so small that it is fine? Also it seems like it was like that before - maybe more noticeable now that the texts are closer together 🤔

andreasphil:
> Looks good! 🙌 I noticed that the H1 is a tiny bit misaligned to the right:
> 
> Image
> Maybe thats so small that it is fine? Also it seems like it was like that before - maybe more noticeable now that the texts are closer together 🤔

I like the attention to detail but that's just the letter sizing of the font, nothing we can do about that 😄

PR #2134, opened 2026-06-19
Risdev 11715 delete and recreate snapshots

ekl176:
This PR enables the job to create the document snapshots to also manage the snapshot lifecycle. 

It will try to recreate a new snapshot after 24 hours (todo: parameterize), no matter how often it runs during that time frame. 
It will also check for actual changes before creating a snapshot. If no changes are detected the old snapshot will be kept in place.
In case a document deletion is recognized it will delete the snapshot right away and start the process to recreate a snapshot.


backend/src/main/java/de/bund/digitalservice/ris/search/service/BulkExportJob.java
```diff
@@ -0,0 +1,89 @@
...
+    this.exportService = service;
+    this.portalBucket = portalBucket;
+    this.clock = clock;
+    this.documentType = documentType;
+    this.changelogService = changelogService;
+  }
+
+  @Override
+  public ReturnCode runJob() {
+    Instant timestamp = clock.instant();
```
andrew-cenkner-digitalservice: Is there a reason we don't use Instant.now() here?
We use Instant.now() in other locations it would simplify the code (constructor, javadoc, bean creation, etc.).
I'm also concerned about code like `Clock.systemDefaultZone()`, because it's usually better to not manage time zones ourself. Instant.now() already works regardless of what time zone you are in, because it stores the offset internally. You can see that by using Instant.now().toString() (The "Z" at the end means UTC+0) and it will correctly evaluate this to the "real" now, regardless of your time zone.

andrew-cenkner-digitalservice (dismissed):

backend/src/main/java/de/bund/digitalservice/ris/search/service/BulkExportJob.java
```diff
@@ -0,0 +1,89 @@
...
+          changes.getChanged().isEmpty()
+              && changes.getDeleted().isEmpty()
+              && !changes.isChangeAll();
+
+      if (noChanges) {
+        logger.info("No new changes detected. Keeping the snapshot");
+        return ReturnCode.SUCCESS;
+      }
+
+      boolean deletionDetected = changes.getDeleted().stream().anyMatch(d -> d.endsWith(".xml"));
```
andrew-cenkner-digitalservice: A `{"change_all" : true}` is also something we should act on right away since it implicitly potentially includes deletions.

PR #2147, opened 2026-06-22
RISDEV-10951 unify spacings throughout the Portal

andreasphil:
- lots spacing and typography adjustments across the application
- introduce `label2` typography variant; rename `label` variant to `label1`
- reorder sorting and pagination controls in advanced search to match simple search

### DIFF
--- a/frontend/src/assets/typography.css
+++ b/frontend/src/assets/typography.css
@@ -38,12 +38,20 @@
   @apply ris-subhead-bold 2xl:ris-heading3-bold;
 }
 
-@utility typo-label-regular {
+@utility typo-label1-regular {
   @apply ris-label2-regular sm:ris-label1-regular 2xl:text-[1.25rem];
 }
 
-@utility typo-label-bold {
+@utility typo-label1-bold {
   @apply ris-label2-bold sm:ris-label1-bold 2xl:text-[1.25rem];
+}
+
+@utility typo-label2-regular {
+  @apply ris-label2-regular 2xl:text-[1.125rem];
+}
+
+@utility typo-label2-bold {
+  @apply ris-label2-bold 2xl:text-[1.125rem];
 }
 
 @utility typo-mono {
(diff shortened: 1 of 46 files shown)


andreasphil:
@UrsKahmann I think the most interesting change is this one: https://github.com/digitalservicebund/ris-search/pull/2147/changes/43a341feae94c23bd873484a9ef1b9e48820fa1d, tweaking the breadcrumbs layout so it

- doesn't shift when it contains an action menu
- actually has `16px` or `24px` spacing (was `24px` or `32px` because the breadcrumbs need some padding for the outline)

I used a similar approach as I did with the back to top button (`data-back-to-top-adjust`)---basically introducing a data attribute that I can select for in the parent component and then adjust styling of the parent component accordingly. It works and it's only a tiny bit of CSS. On the other hand, it's maybe a bit hard to follow if you encounter it in the codebase and don't know what it's about? Not sure...

UrsKahmann:
> @UrsKahmann I think the most interesting change is this one: [43a341f](https://github.com/digitalservicebund/ris-search/commit/43a341feae94c23bd873484a9ef1b9e48820fa1d), tweaking the breadcrumbs layout so it
> 
>     * doesn't shift when it contains an action menu
> 
>     * actually has `16px` or `24px` spacing (was `24px` or `32px` because the breadcrumbs need some padding for the outline)
> 
> 
> I used a similar approach as I did with the back to top button (`data-back-to-top-adjust`)---basically introducing a data attribute that I can select for in the parent component and then adjust styling of the parent component accordingly. It works and it's only a tiny bit of CSS. On the other hand, it's maybe a bit hard to follow if you encounter it in the codebase and don't know what it's about? Not sure...

Ahh good catch, I guess I didn't calculate in the padding when setting the vertical spacing. I think the solution with the data attribute is fine 👍  I think a comment on the action menu where the data attribute is set would be helpful. So when reading that part its easier to understand that the attribute is used by the breadcrumbs for spacing adjustments... But just a suggestion :) 

Otherwise looking really good. Two tiny things I noticed, but I think they have been like that before:
1. On the english translations page there is a scrollbar on the title 🤔 
grafik

2. The translation pages are missing the correct bottom spacing. But that's actually on me I guess 😓 

So all good, if you don't mind, it would be great if you could fix those two things as they fit into this PR topic wise. 🫶

UrsKahmann (approved):

PR #2148, opened 2026-06-23
RISDEV-10827 Extract tabs layout into dedicated component

UrsKahmann:
- refactor tab layout usage by extracting it into a dedicated component
- reuse the `TabsLayout` component in all places where tabs are needed
- had to add vertical spacing around the translations page tab content -> this will probably be updated with the unify spacing PR

RISDEV-10827

### DIFF
--- a/frontend/src/components/TabsLayout.vue
+++ b/frontend/src/components/TabsLayout.vue
@@ -1,0 +1,56 @@
+<script setup lang="ts">
+import { Tab, TabList, Tabs } from "primevue";
+import { NuxtLink } from "#components";
+
+export type TabView = {
+  label: string;
+  path: string;
+  icon: Component;
+  analyticsId?: string;
+};
+
+const { views } = defineProps<{
+  views: OneOrMore<TabView>;
+}>();
+
+const route = useRoute();
+
+const currentView = computed(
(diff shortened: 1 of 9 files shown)


frontend/src/components/TabsLayout.vue
```diff
@@ -0,0 +1,56 @@
...
+
+const currentView = computed(
+  () => route.query.view?.toString() ?? views[0].path,
+);
+</script>
+
+<template>
+  <div>
+    <div class="border-b border-gray-400">
+      <nav class="wrapper -mb-1 overflow-x-auto pt-1" aria-label="Tab">
```
UrsKahmann, line 26: @andreasphil for some reason the to shadow/border of each tab was not showing anymore so I added the `pt-1`. I actually didn't quite get what changed as I didn't do any layout adjustments... I wasn't sure if this "fix" is okay or if it covers up a different problem, let me know what you think.

andreasphil: Probably because of the extra wrapping container that we have now. Let's not overthink it as long as it works, this will no longer be relevant once we implement the redesigned tabs (and I don't like the implementation with the shadows anyway, it has a bunch of issues)

andreasphil (approved):

PR #2153, opened 2026-06-23
RISDEV-0000 resolve some code smells reported by Sonar

andreasphil:


UrsKahmann (commented):
Thanks for taking care 🙏 👍

PR #2162, opened 2026-06-26
RISDEV-10395 hide text content of search teaser sections if the match is only in the name

andreasphil:


frontend/src/components/search/CaseLawSearchResult.spec.ts
```diff
@@ -169,21 +169,41 @@ describe("CaselawSearchResult", () => {
...
+      },
+      {
+        field: "decisionGrounds",
+        value: "Entscheidungsgründe.",
+        text: "<mark>Entscheidungsgründe</mark>.",
+      },
     ];
 
-    fields.forEach(({ field, value }) => {
+    fields.forEach(({ field, value, text }) => {
```
UrsKahmann, line 201: **suggestion:** Use [`it.each(...)`](https://vitest.dev/api/test.html#test-each) instead of putting the test inside a for-loop.

UrsKahmann (commented):
Looks good 👍

PR #2174, opened 2026-06-29
RISDEV-11882 align pages to grid layout

andreasphil:
This is the first batch of changes to align the Portal frontend pages to a common grid:

- Introduces a new `base-grid` utility that sets common responsive properties
- Aligns the static content, document, searches and landing pages to the grid


UrsKahmann (commented):
👍  thanks for also cleaning up the default slot usage 🙌

PR #2179, opened 2026-07-01
RISDEV-11789 Use ris-abbreviation as fallback

UrsKahmann:
## This PR:
- implements the fallback logic for the norms abbreviation
- adds basic tests and adapts other backend tests where needed (test will be refactored in the near future again when the TestDataBuilder is implemented)
- updates e2e test data
- updates openapi.json
- renames `officialAbbreviation` properties to `abbreviation` except for the opensearch field -> renaming would make it more difficult to run a reindex without issues -> opensearch field renaming will be done in a separate PR


backend/src/main/java/de/bund/digitalservice/ris/search/mapper/NormLdmlToOpenSearchMapper.java
```diff
@@ -278,22 +279,21 @@ public static String parseShortTitle(String original) {
...
+   * doesn't exist uses the ris:abkuerzung as fallback.
+   *
+   * @param xmlDocument
+   * @return official abbreviation or ris:abkuerzung (fallback)
+   */
+  private static Optional<String> getAbbreviation(XmlDocument xmlDocument) {
+    Optional<String> officialAbbreviation =
+        xmlDocument.getNonEmptyElementByXpath(X_PATH_SHORT_TITLE_ABBREVIATION);
+    Optional<String> risAbbreviation = xmlDocument.getNonEmptyElementByXpath(X_PATH_RIS_ABKUERZUNG);
+
```
ekl176, line 292: some of these documents can be quite large so we might not want to parse the fallback in advance, but only when we actually need it.

ekl176 (approved):

PR #2181, opened 2026-07-02
RISDEV-0000 general frontend cleanup

andreasphil:
- reduces amount of files by removing unnecessary components and merging files
- simplify UI code by changing many many manual imports to auto-imports
- cleaner file names and auto import names by removing unnecessary prefixes
- improve unit test reliability by removing unnecessary stubbing
- communicate intended use of components more clearly by moving single-purpose components into their subfolders


UrsKahmann (commented):
Awesome, thank you very much for cleaning things up 🧹 🚀

PR #2191, opened 2026-07-03
RISDEV-10394 review feedback fixes to grid alignment

andreasphil:
Fixes remaining functional review findings:

- removes explicit font size for printing
- limit text content to 7 columns on xl breakpoint
- translation search results always span the full page width


UrsKahmann (commented):
👍

PR #2213, opened 2026-07-07
RISDEV-0000 always show full norm heading

andreasphil:
Removes truncation for norm headings so it's always fully visible (even if the heading is really long). Also removes the now-unused `ExpandableText` component.


UrsKahmann (commented):
Nice 🚀 🗑️

PR #2214, opened 2026-07-07
Implementation of use of xsd schema documentation

rvp-c:
Adds an XSD-driven documentation pipeline for the backend: OpenAPI field descriptions are now parsed directly from the XSD schema's <xs:documentation> text instead of being hand-written in @Schema annotations, so the docs stay in sync with the actual data format.

Core pieces: XSDDescriptionParser parses the caselaw/adm/literature XSDs into descriptions, a new @RISSchema annotation marks fields to look up, and XSDPropertyCustomizer injects the matched text at spec-generation time (falling back to a manual description if none is found). Applied so far to a first slice of CaseLawSchema fields, not the whole schema yet.

Where to see it: generated tables in doc/readme/{caselaw,adm,literature}.md (regenerate via ./gradlew xsdDocumentation), or /v3/api-docs//swagger-ui.html on a running backend.


.github/workflows/xsd-documentation.yml
```diff
@@ -0,0 +1,51 @@
+name: "Generate documentation for XSD Schema documentation keys"
+
+on:
+#  push:
+#    branches: [main]
+#    paths:
+#      - "backend/build.gradle.kts"
```
malte-laukoetter, line 7: can this be removed?

backend/src/main/java/de/bund/digitalservice/ris/search/xsd/DocumentationElement.java
```diff
@@ -0,0 +1,80 @@
...
+        stringBuilder.append("<li>").append(line.trim().substring(1).trim()).append("</li>");
+      } else {
+        if (listOpen) {
+          stringBuilder.append("</ul>");
+        }
+        listOpen = false;
+        stringBuilder.append(line.trim()).append("<br>");
+      }
+    }
+    return stringBuilder.toString();
```
malte-laukoetter, line 48: should we remove the tailing `<br>` tag?

malte-laukoetter (approved):

andrew-cenkner-digitalservice (changes requested):
The intention of the PR is good, but I think this could be better served by two smaller PRs and not just to make the review easier.

The first PR would actually be to the ris-xml-schema repository including MarkdownGenerator and XSDDescriptionParser and would be responsible for generating the markdown files. The reason the markdown file generation should live in the ris-xml-schema is because the markdown files change if and only if the xsd files change. The markdown files are just a subset of the data in the xsd files.
This has 2 advantages. First, if a change to xsds breaks md generation, the author knows immediately. Second, the separation of ownership is more clear. Ownership of two representations of the same data should not be split.

The second PR would be the parts that didn't get added to ris-xml-schema in order for ris-search to use the new @RISSchema annotation. ris-search already depends on ris-xml-schema, so splitting into 2 PRs won't cause any code duplication.

PR #2216, opened 2026-07-08
RISDEV-11960: Use builder for norm test data

UrsKahmann:
## This PR:
- makes use of the `NormTestDataBuilder` to create the test data for the `NormLdmlToOpensearchMapper`
- adapts the `NormTestDataBuilder` where needed
- adapts tests where needed
- renames the `ris:gegenstandslos` field to correctly match the sepcification
- refactors tests to use `assertJ` assertions
- removes all obsolete raw xml test files
- adds an ADR with the reasons for the decision to switch to the builder approach


andrew-cenkner-digitalservice (commented):
Approved

PR #2233, opened 2026-07-13
RISDEV-11519 adjust visible titles for court decisions and legislation

andreasphil:
- exposes the Titelzeile (headline) of court decisions in the JSON API
- adds highlighting to caselaw headline and legislation short title
- fixes legislation short title API types so that it can be `null`
- updates caselaw and legislation search results to show the new titles
- shows headline on caselaw detail page


backend/src/test/java/de/bund/digitalservice/ris/search/unit/mapper/CaseLawLdmlToOpenSearchMapperTest.java
```diff
@@ -57,4 +58,15 @@ void shouldMapToCaseLawDocumentationUnitCorrectly() {
             "ensuing decision file number, ensuing decision court type",
             "ensuing decision file number, ensuing decision court type");
   }
+
+  @Test
```
andreasphil, line 62: Feedback wanted: Is it okay to test this like this? I didn't find a way of doing it with the templates, but maybe I'm missing something.

UrsKahmann: You could add yet another if block in the template to allow hiding the short title.. But I don't think its worth it. This is the exact kind of test case I was struggling in the norm context with as well as the template doesn't give enough flexibility.. I will push for using the builder approach for caselaw test data as well.. 
That said: I think your approach is fine for now, at least I don't see a better way of doing it with the current setup.

frontend/src/components/search/SearchResultHeader.vue
```diff
@@ -5,28 +5,35 @@ export interface SearchResultHeaderItem {
   value: string;
 }
 
-const { icon, items } = defineProps<{
+const { icon, items, secondaryItem } = defineProps<{
   icon?: Component;
   items: SearchResultHeaderItem[];
+  secondaryItem?: string;
```
UrsKahmann, line 11: **nitpick:** I find the name secondary**Item** a bit misleading. Because the `items` prop is of type `SearchResultHeaderItem[]` I would expect the `secondaryItem` to be of type `SearchResultHeaderItem`. Maybe something like `secondaryHeadline` or something would be more precise? But also fine to leave it like it is ✌️

UrsKahmann (commented):
Really really nice work 🥇

frontend/src/components/search/SearchResultHeader.vue
andreasphil: Fixed in #2238

andrew-cenkner-digitalservice (commented):
Approved

PR #2236, opened 2026-07-15
RISDEV-0000 pipeline tweaks

andreasphil:
Fixes several findings reported by Zizmor:

- stop persisting credentials when checking out the repository (reduces the risk of credentials accidentally being included in downloadable artifacts)
- remove shell script interpolation in favor of env variables (those are escaped, reduced risk of shell injections)
- more fine grained permissions and secrets inheritance
- formatting
- check for Dependabot as an actor in a way that can't easily be spoofed


UrsKahmann (approved):
Looks good, thanks for the improvements 🔒

PR #2240, opened 2026-07-16
RISDEV-7898 add norm abbreviation route

UrsKahmann:
- add a page route that allows to open a norm directly via an abbreviation

RISDEV-7898


frontend/src/middleware/normsAbbreviationRedirect.ts
```diff
@@ -0,0 +1,23 @@
+import type {
```
andreasphil, line 1: nitpick: should be named "normsAbbreviationRedirect.ts" to be consistent with spelling of other file names (the middleware [will still be named](https://nuxt.com/docs/4.x/directory-structure/app/middleware#:~:text=Name%20of%20middleware%20are%20normalized%20to%20kebab%2Dcase) "norms-abbreviation-redirect")

andreasphil (commented):
Just some nitpick about naming, otherwise LGTM 👍

PR #2241, opened 2026-07-17
RISDEV-0000 interrupt page setup when data can't be loaded

andreasphil:
Fixes an issue where the page setup would continue to run in the background, even though required data could not be loaded. This PR changes the error handling so it throws an error instead of asynchronously redirecting to an error page, ensuring the setup is interrupted immediately and nothing silently crashes in the background.


UrsKahmann (commented):
👍 🚀

PR #2242, opened 2026-07-17
RISDEV-11566 rename content page paths to german

UrsKahmann:
- first part of RISDEV-11566: renaming the content page paths to german
- the document and search paths will be done separately to reduce the amount of changes in one PR

RISDEV-11566


andreasphil (commented):
👍

PR #2243, opened 2026-07-17
RISDEV-11566 rename search pages paths to german

UrsKahmann:
- second part of RISDEV-11566: renaming the search pages paths to german

RISDEV-11566


andreasphil (commented):
👍

PR #2262, opened 2026-07-21
RISDEV-11566 german urls document pages

UrsKahmann:
- last part of the ticket, changing the document page urls to use german names

RISDEV-11566


andreasphil (commented):
👍

PR #2263, opened 2026-07-21
risdev-12046-query-work-examples-from-repo

ekl176:
This PR changes how we retrieve all expressions for a specific work. Instead of manually querying doc_value fields we use the repository to retrieve expressions. 
This makes sure that all fields we retrieve are in the same format, since keyword and text fields might be normalized differently.
To minimize the performance impact we exclude artice_texts and article_names on query time. 
The next index recreation will also remove it from the index _source to generally improve storage usage and query time.


UrsKahmann (approved):
Looks good to me, but I have little knowledge about the potential performance impacts. Lets just try it out I guess 🚀

PR #2265, opened 2026-07-21
RISDEV-11576 prototype navigating article versions via eId

andreasphil:
Technical exploration of linking versions of the same article in past/future versions of a norm via their eId. Supports listing all versions of a norm containing elements with that eId, and grouping by validity time period of the article:

Bildschirmfoto 2026-07-21 um 16 48 14

Also see [Copilot session dump](https://github.com/digitalservicebund/ris-search/blob/1a111f67f31ab1eba56e733213e3cbfd5184ab7b/doc/prototypes/norm-article-version-navigation.md) and an alternative approach using Dokumentnummern: #2267


andreasphil:

PR #2266, opened 2026-07-21
RISDEV-12052 use BundesSans font for API docs

UrsKahmann:
RISDEV-12052


andreasphil (approved):
looks good, thank you for the changes and for adding the formatting and linting!

PR #2267, opened 2026-07-21
RISDEV-11576 prototype navigating article versions via Dokumentnummer

andreasphil:
Technical exploration of linking versions of the same article in past/future versions of a norm via their Dokumentnummer. Supports listing all versions of a norm containing elements with that Dokumentnummer, and grouping by validity time period of the article:

Bildschirmfoto 2026-07-21 um 17 48 13

Also see [Copilot session dump](https://github.com/digitalservicebund/ris-search/blob/d71dc4a8a9847e0dce6ec9c8cfa634e75c23ee6d/doc/prototypes/norm-article-version-navigation.md) and an alternative approach using eIds: #2265


andreasphil:

PR #2287, opened 2026-07-27
RISDEV-0000 pin brace-expansion to circumvent version with CVE

UrsKahmann:
RISDEV-0000


andreasphil (commented):
Thanks for taking care of this 👍

PR #2289, opened 2026-07-27
RISDEV-7898 use ris-abbreviation for speakable url

UrsKahmann:
- adapt middle ware to use ris-abbreviation for redirect
- add ris-abbreviation field to `LegislationExpressionSchema`
- use ris-abbreviation to build the speakable url for norms 

RISDEV-7898

---
**Other**:
- exclude `generated-api-types.d.ts` from formatting and linting


frontend/src/components/documents/actionMenu/NormActionMenu.vue
```diff
@@ -9,24 +9,36 @@ import { useXmlActionItem } from "~/composables/useActionMenuItem/useXmlActionIt
 import type { LegislationExpression } from "~/types/api";
 
 const { metadata, translationUrl } = defineProps<{
-  metadata: LegislationExpression | undefined;
+  metadata: LegislationExpression;
   translationUrl: string | undefined;
 }>();
 
+function normalizeRisAbbreviation(risAbbreviation: string) {
```
andreasphil, line 16: question: what happened to the discussion we had about allowing more special characters? Is this being tracked in a different ticket?

andreasphil (commented):
just a question about scope, looks good otherwise :)

frontend/src/components/documents/actionMenu/NormActionMenu.vue
UrsKahmann: Correct, this is just the first iteration. Actually we just got the mapping table for the other special chars -> this is a second ticket I will do next :)

PR #2290, opened 2026-07-27
RISDEV-10550 fix visual issues in mobile overlay

andreasphil:
- removes wrong whitespace on first render in mobile Safari
- overrides PrimeVue's buggy autocomplete overlay positioning logic

---

(the issue is that PrimeVue's autocomplete positioning logic somehow doesn't consider margins and paddings, leading to wrong absolute positioning of the overlay. Adding a patch that overrides the logic. I know it's ugly, but I couldn't find a better way, and we're likely to remove PrimeVue soon anyway.)

### DIFF
--- a/frontend/e2e/suche.spec.ts
+++ b/frontend/e2e/suche.spec.ts
@@ -1195,6 +1195,45 @@ test.describe("mobile filter and sort drawers", () => {
     );
   });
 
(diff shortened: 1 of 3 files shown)


UrsKahmann (commented):
👍

PR #2291, opened 2026-07-27
RISDEV-0000 disable insecure request upgrading during development

andreasphil:
This is ignored by Chrome anyway and makes the page unusable in Safari. Disabling it during development.


UrsKahmann (commented):
👍

PR #2292, opened 2026-07-29
RISDEV-11960 implement review feedback part 1

UrsKahmann:
- refactor entities to use simple constructors instead of builder pattern if they don't have more than 2-3 parameters

RISDEV-11960


andrew-cenkner-digitalservice (commented):
Approved

PR #2297, opened 2026-07-29
RISDEV-0000 try to fix source map uploading

UrsKahmann:
RISDEV-0000


andreasphil (approved):
Nice, thank you! Let's see if this helps 🤞

PR #2299, opened 2026-07-29
Risdev 12103 add missing cl fields

MagdaN:
- Adds a /rechtsprechung API endpoint that exposes case law fields under their German property names, adding fields that had been missing until now, mirroring the existing  /case-law endpoint
  - Adds field aliases (with test coverage) for that endpoint 
  - Integrates the ris-xml-schema XSD schemas as a test dependency and adds a CaseLawXmlValidator test utility (mirroring the   existing NormXmlValidator/LiteratureXmlValidator) that validates case law LDML against the official decision/pending-proceeding   schemas
  - Wires that validation into the case-law test fixture template rendering (so any future fixture change is checked against the  schema automatically) and into the LdmlTestDataSchemaConformanceTest sweep over real e2e test data
  - Fixes real schema violations this surfaced in 13 e2e-data case law fixture files
  - Makes the case-law test fixture template fully bidirectional for the "Entscheidung" vs. "Anhängiges Verfahren" (pending  proceeding) variants, hiding decision-only fields when pendingProceeding=true and vice versa


backend/src/main/java/de/bund/digitalservice/ris/search/models/ldml/caselaw/AknDissentingOpinion.java
```diff
@@ -0,0 +1,61 @@
...
+  @XmlElement(name = "block", namespace = AKN_NS)
+  private AknMotivationBlock aknMotivationBlock;
+
+  /**
+   * Formats paragraphs and opinion blocks into a single comma-separated string.
+   *
+   * @param risMeta the RIS metadata used to resolve opinion authors' display names, may be null
+   * @return the formatted text combining paragraphs and opinions
+   */
+  public String toFormattedText(RisMeta risMeta) {
```
UrsKahmann, line 40: **question:** There are a few cases like this where the model class has a function that actually contains quite a bit of logic. If I didn't miss anything these are not well covered by tests.. I think it would be good to add unit tests for those cases? 
(Also for other model classes that have functions with more complex logic)

UrsKahmann (commented):
Thank you for tackling this!! 🚀 
In general this looks good, but I didn't check every detail, its simply too much 🫣 
A few thoughts:
- I think I haven't fully understood what the new endpoint is for? Is that just temporary? Or will this stay and we will have a german and an english version for caselaw endpoints? That I would find a bit odd 🤔 
- Out of scope for this PR but: In my opinion the caselaw test data template is really getting out of hand 😅 The templates are huge and with all the if else statements its super hard to grasp what the final result will actually look like. In my opinion we should think about using the TestDataBuilder appraoch for caselaw (doing this for norms already). It should be even easier as for caselaw the model classes already exist :)

ekl176 (commented):
Could we split this up in separate PRs and merge the opensearch changes first? 
Once the new fields are indexed they should also be found using the existing case-law endpoint. 

We might also need to add the new fields to the CaseLaw highlighter if we want those fields to be shown in the preview in the advanced search.

MagdaN:
@UrsKahmann the idea was to create the new endpoint first, then switch the frontend to use this endpoint and then remove the old endpoint. Yes, I agree the templates are too big and I was going to use TestDataBuilder approach for Caselaw as well but had to stop at some point as the PR is already quite big.

PR #2300, opened 2026-07-30
RISDEV-11065 show validity end date in search result header

UrsKahmann:
- in none prototype environment: show validity  from - to date in search result header if a "to" date exists

RISDEV-11065

### DIFF
--- a/frontend/src/components/search/NormSearchResult.vue
+++ b/frontend/src/components/search/NormSearchResult.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import type { Dayjs } from "dayjs";
 import IcBaselineBalance from "~icons/ic/baseline-balance";
 import type { RouteLocationRaw, RouteLocationAsPath } from "#vue-router";
 import type { SearchResultHeaderItem } from "~/components/search/SearchResultHeader.vue";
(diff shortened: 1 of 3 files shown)


frontend/src/components/search/NormSearchResult.vue
```diff
@@ -31,20 +30,23 @@ const secondaryTitle = computed<SearchResultHeaderItem | undefined>(() =>
...
+  let dateValue: string | undefined = dateFormattedDDMMYYYY(
+    searchResult.item.exampleOfWork.legislationDate,
+  );
 
   if (privateFeaturesEnabled) {
     const coverage = temporalCoverageToValidityInterval(
       searchResult.item.temporalCoverage,
     );
-    date = coverage?.from;
+    const from = dateFormattedDDMMYYYY(coverage?.from);
```
andreasphil, line 41: question: does the `from` date always exist? The optional chaining here suggests it doesn't. If that is the case then the `dateValue` formatting below could end up as `undefined`, leading to an item with `{ value: undefined }`. It took me a while to understand that this would be filtered out below, doesn't feel very clean to me to first "clutter up" the items and filter them again later...

since you're only adding to the existing logic feel free to leave it though if you don't want to change it

andreasphil (commented):
👍

frontend/src/components/search/NormSearchResult.vue
UrsKahmann: Yeah good point, I am not super sure if the `from` does always exit. I think the initial thought why it is setup this way was that the abbreviation and the date could potentially be null. And having multiple `if (not not nulll) then add to the array` made it harder to see at one glimpse what values are potentially shown and in what order.. All tough I guess this is a weak argument if its only two values that are potentially undefined 😅 
The abbreviation will also become an always existing value (because the fallback ris-abbreviation now always exists, I stil need to do a refactor in the backend).. And then only the date could be null. I agree after that it would probably be cleaner to get rid of the filtering and only add the date value at the end if not undefined... Long story short: I will change it, thaaaanks :D

PR #2303, opened 2026-07-31
RISDEV-11721 apply Sentry default error reporting to more error sources

andreasphil:
This PR adds manual error filtering for 3xx-4xx error codes thrown via `showError` and `createError`. This matches the default behavior of the Sentry SDK on the server and client side for errors thrown via other channels.

See Sentry SDK for [client](https://github.com/getsentry/sentry-javascript/blob/61b7d52d721e181f86f30620928ef0fb39cdf42f/packages/nuxt/src/runtime/plugins/sentry.client.ts#L71) and [server](https://github.com/getsentry/sentry-javascript/blob/61b7d52d721e181f86f30620928ef0fb39cdf42f/packages/nuxt/src/runtime/hooks/captureErrorHook.ts#L24)

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -7,5 +7,12 @@ if (config.public.sentryDSN) {
     dsn: config.public.sentryDSN,
     environment: config.public.sentryEnvironment,
     tracesSampleRate: 1,
+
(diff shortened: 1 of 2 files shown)


UrsKahmann (commented):
Niceee 🚀

PR #2304, opened 2026-07-31
RISDEV-11454 drop chunk loading errors

andreasphil:
Client-side chunk loading errors (e.g. after a deployment invalidates old hashed chunk URLs) are already handled by Nuxt's `nuxt:chunk-reload` plugin, which performs a hard reload. Sentry reports them as unhandles errors even though the user is not affected. Drop them as noise.

See https://nuxt.com/docs/4.x/getting-started/error-handling#errors-with-js-chunks


UrsKahmann:
Nice 👍

PR #2305, opened 2026-07-31
RISDEV-9129 hide text tab for empty norms

UrsKahmann:
- hide text tab for norms that don't have any body content (specifically no contentn in the `div` with `class="akn-body"`)
- the logic for caselaw will be done in a separate PR

RISDEV-9129


andreasphil (approved):
👍

PR #2307, opened 2026-07-31
RISDEV-9129 hide tablist for empty caselaw documents

UrsKahmann:
- hide the tablist and only show the details if the caselaw document is considered empty
- does not yet add the info box for "Vorabdokumente", this will be done separately because a new index field is needed

RISDEV-9129


andreasphil (commented):
👍

PR #2319, opened 2026-08-03
RISDEV-0000 cleanup show error

andreasphil:
- removes almost all instances of `showError` in favor of `throw createError` for more consistent and predictable error handling
- only exception: kept `showError` inside a watcher, which behave differently due to reactivity
- moved the error handling of the norm versions composable to the view instead of handling it internally


UrsKahmann (commented):
Nicee 🚀

PR #2320, opened 2026-08-03
RISDEV-10317 Alert on failed cronjobs

andrew-cenkner-digitalservice:
*Revert logging change
*Put alerting rule and markdown under version control *Remove dead code

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexSyncJob.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/service/IndexSyncJob.java
@@ -9,7 +9,6 @@ import java.util.List;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.message.StringMapMessage;
 import org.springframework.scheduling.annotation.Async;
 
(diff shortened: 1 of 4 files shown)


UrsKahmann (commented):
Thanks Andrew, nice work! 
I wonder if ris-search is actually the right place, I think I'd prefer to put the grafana alert definitions and documentation into the infra repository as its more like "platform" configuration in my opinion 🤔 Wdyt? 

Otherwise it looks, good (from what I understand). I think its a great first step and the 8h interval also seems reasonable. Ideally in the future we can define the alerts via terraform or kubernetes but I guess we will have to wait for the platform team on this.

andrew-cenkner-digitalservice:
> Thanks Andrew, nice work! I wonder if ris-search is actually the right place, I think I'd prefer to put the grafana alert definitions and documentation into the infra repository as its more like "platform" configuration in my opinion 🤔 Wdyt?
> 
> Otherwise it looks, good (from what I understand). I think its a great first step and the 8h interval also seems reasonable. Ideally in the future we can define the alerts via terraform or kubernetes but I guess we will have to wait for the platform team on this.

Yes to both. We can move this to the infra repo when the platform team enables alerts via terraform.

PR #2323, opened 2026-08-05
RISDEV-12108 hide empty datafields

UrsKahmann:
- refactor `DetailsList` component
    - define detail list items as input
    - only display entries with values
- simplifies and unifies usage across documents

RISDEV-12108


frontend/src/components/documents/DetailsList.spec.ts
```diff
@@ -1,10 +1,178 @@
...
+
+const iconStub = { template: "<span />" };
+
+function renderComponent(items: DetailsListItem[]) {
+  return render(DetailsList, {
+    props: { items },
+    global: {
+      stubs: {
+        NuxtLink: nuxtLinkStub,
+        IcOutlineFileDownload: iconStub,
```
andreasphil, line 19: question: why do we need to stub the icon?

frontend/src/components/documents/DetailsList.vue
```diff
@@ -1,5 +1,110 @@
...
+          {{ value }}
+        </dd>
+      </template>
+
+      <dd
+        v-else-if="item.type === 'html'"
+        class="typo-label1-regular col-span-12 md:col-span-9 md:col-start-4 lg:col-span-6 lg:col-start-4"
+        :class="item.htmlClass"
+      >
+        <!-- eslint-disable-next-line vue/no-v-html -->
```
andreasphil, line 90: todo: remove, we don't use eslint anymore

frontend/src/components/documents/DetailsList.vue
```diff
@@ -1,5 +1,110 @@
...
+        <div v-html="item.html" />
```
andreasphil, line 91: question: how is sanitizing working in this case? Does the HTML get sanitized?

andreasphil (commented):
nice refactor! I had a few remarks :)

frontend/src/components/documents/DetailsList.spec.ts
UrsKahmann: Good catch, we don't 🫣  I will remove it!

frontend/src/components/documents/DetailsList.vue
UrsKahmann: Ahh right. It wasn't sanitized before e.g. previously: 
```
 <DocumentsDetailsListEntry label="Fußnoten:">
    <template v-if="htmlParts.headingNotes">
      <div class="footnotes" v-html="htmlParts.headingNotes" />
    </template>
  </DocumentsDetailsListEntry>
```
But you are right, it probably should get sanitized.

PR #2325, opened 2026-08-05
RISDEV-0000 minor cleanup

andreasphil:
- merges the German and English codes of conduct into one file (only one is recognized by GitHub anyway)
- renames the licence checker script to camel case, consistent with the other filenames on that folder level
- removes the unnecessary root level `run.sh` script
- moves the lombok config to the `backend` folder, which is the only place where it's used


UrsKahmann (commented):
Thanks for cleaning this up 🙏 ✅

PR #2328, opened 2026-08-06
RISDEV-12201, RISDEV-12204 migrate button and progress spinner components

andreasphil:


frontend/src/components/ui/Button.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    render(Button, { props: { label: "x", disabled: true } });
+
+    expect(screen.getByRole("button")).toBeDisabled();
+  });
+
+  it("emits click events", async () => {
+    const onClick = vi.fn();
+    render(Button, { props: { label: "x" }, attrs: { onClick } });
+
+    await fireEvent.click(screen.getByRole("button"));
```
UrsKahmann, line 64: **question:** Why use `fireEvent` here instead of `user.click` with the `userEvent`? I saw that we mix this a bit in our tests but mostly use the `userEvent.setup()` and `user.<some action>`. Also the docs for [userEvent](https://testing-library.com/docs/user-event/intro#differences-from-fireevent)  sound like that in cases where the `userEvent` can be used it should be preferred over [fireEvent](https://testing-library.com/docs/dom-testing-library/api-events/#fireevent).

UrsKahmann (commented):
Looks good, just one comment with a small question 👍

PR #2335, opened 2026-08-07
RISDEV-11432 Infra and Backend Maintenance tasks

andrew-cenkner-digitalservice:
* Deleted old config and mentions to postgres database no longer used by portal
* Refactored spring profile usage so that "dev" now represents our local profile enabling a more clean separation that application.yaml is used for properties common to all profiles.
* Minor refactor of a few properties to test the concept will work.


backend/src/main/java/de/bund/digitalservice/ris/search/config/CorsConfig.java
```diff
@@ -72,7 +72,7 @@ public void addCorsMappings(CorsRegistry registry) {
    * @return The web mvc configurer
    */
   @Bean
-  @Profile({"e2e", "prototype", "default"})
+  @Profile({"prototype", "dev"})
```
UrsKahmann, line 75: **question:** Was the e2e profile removed here on purpose?

.github/workflows/pipeline-e2e.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Start backend locally (background)
         working-directory: ./backend
         env:
-          SPRING_PROFILES_ACTIVE: e2e,default
+          SPRING_PROFILES_ACTIVE: dev,e2e
```
UrsKahmann, line 79: **question:** Is `dev` even needed here anymore? If I see it correctly, the `dev` profile only specifies the `server.port` and the `e2e` profile has the same definition, so that should be enough right?

backend/src/main/resources/application.yaml
```diff
@@ -1,20 +1,10 @@
...
-      minimum-idle: 20
-      validation-timeout: 3000
-      idle-timeout: 60000
+  profiles:
+    # all environments override this property, so it's only used locally.
+    default: dev
 
 server:
-  port: 8090
+  port: 8080
```
UrsKahmann, line 7: **question:** I am a bit confused and unsure if this will work: 
- the application runs on port 8080 now 
- the docker image exposes port 8090
- but then in our infra repo we use port 8080 again 🤔  
 
And this has been working so far but I am unsure if it shouldn't all be the same port?

UrsKahmann (commented):
Thanks for cleaning this up, left a few comments 🙂

backend/src/main/java/de/bund/digitalservice/ris/search/config/CorsConfig.java
andrew-cenkner-digitalservice: Yes, because the e2e "run" is different than the e2e spring profile. The run starts with 2 profiles : dev & e2e so it was redundant. I put a daily discussion topic for this topic.

.github/workflows/pipeline-e2e.yml
andrew-cenkner-digitalservice: I should have commented that this PR is only step 1 of the story. I needed to merge to main to make sure my changes work as expected. Next steps are to move things from application.yaml to dev.yaml so it will be needed, but isn't really needed yet.

backend/src/main/resources/application.yaml
andrew-cenkner-digitalservice: This was already in an unclear state and this change doesn't change the values, just the way they get assigned. dev and e2e used 8090. The other 4 (stackit profiles) used 8080.
The change made the default follow the stackit profiles, and dev & e2e are the "exceptions" so now we only override the port 1 time instead of 4 times.
I would also like to discuss why we use 2 different ports.

PR #2347, opened 2026-08-10
RISDEV-0000 e2e test await url before navigating back

UrsKahmann:
- try to fix two flaky e2e test by awaiting the expected url before navigating back

RISDEV-0000


andreasphil (commented):
👍

PR #2354, opened 2026-08-10
RISDEV-11721 improved client error status detection for Sentry

andreasphil:
- Updates error status filtering to also work with errors that have been serialized in SSR, and therefore lost their Nuxt error type
- Fixes a crash in the PostHog setup in older browsers that don't support the Cookie Store API, which was the underlying issue that caused the error report


UrsKahmann (commented):
👍

PR #2356, opened 2026-08-11
RISDEV-10909 make data table internal cell spacing less dependent on content height

andreasphil:
Instead of deriving the line height from content height + padding, set a minimum line height that matches the designs visually, reduce vertical padding, and center the items. Looks the same, but gives cell content a bit more flexibility before it pushes the size of the row.


UrsKahmann (commented):
✅

PR #2361, opened 2026-08-11
RISDEV-11432 Infra and Backend Maintenance tasks

andrew-cenkner-digitalservice:
* Change dev from using 8090 to 8080 so everything is the same
* Clean up properties to follow the team agreed convention
* Reverted spring.config.activate.on-profile experiment
* Testing removal of uneeded kubernetes config on staging
* Changes e2e to only use 1 profile


UrsKahmann (commented):
✅

PR #2362, opened 2026-08-11
put generated nlex objects to version control

ekl176:


UrsKahmann (commented):
Looks good ✅  Only thing missing: Maybe make sure that the generated files are excluded from sonar, linting, spotless etc?

PR #2364, opened 2026-08-11
RISDEV-12137 replace accordion, radio button, select, and message

andreasphil:
Migrates the following components:

- RISDEV-12199 accordion
- RISDEV-12200 radio button
- RISDEV-12191 select
- RISDEV-12197 message


andreasphil:
@UrsKahmann sorry for the big PR, easiest way to review is probably commit by commit. A lot of it is also just stories and unit tests

frontend/src/components/ui/Accordion.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    expect(screen.getByRole("button")).toHaveTextContent("Show More");
+
+    await rerender({ modelValue: true });
+    expect(screen.getByRole("button")).toHaveTextContent("Show Less");
+
+    await rerender({ modelValue: false });
+    expect(screen.getByRole("button")).toHaveTextContent("Show More");
+  });
+
+  it("toggles when used with v-model", async () => {
```
UrsKahmann, line 82: **question:** Does this test actually add any more value? For me it looks like that what it tests is already covered by the other tests 🤔

UrsKahmann (commented):
Looks really good 🚀

frontend/src/components/ui/Accordion.spec.ts
andreasphil: what I was trying to test here is that we can also manage the toggle state externally, but you're right that this will probably always pass/fail the same as the other tests that check the model behavior, so it's a bit redundant. I'll remove it 👍

andreasphil: sneaked into #2369

PR #2366, opened 2026-08-12
RISDEV-12260 Move Literature html generation to library

andrew-cenkner-digitalservice:
* Extract literature xslt to html library


MagdaN (commented):
Looks good! Thank you!

PR #2369, opened 2026-08-12
RISDEV-12253 render Eingangsformel footnotes in legislation HTML

andreasphil:
Footnotes are not rendered automatically, but need to be "enabled" in the HTML transform per element. This was missing for Eingangsformeln. This PR enables it.

Changing this led to another issue: footnotes from the official TOC (located in the Eingangsformel) were rendered as a sibling of the TOC with no shared container. So rendering more footnotes would confuse the frontend about which footnotes belong to the TOC, and which belong to the Eingangsformel. This PR puts the TOC and it's footnotes into a shared container to distinguish them from regular Eingangsformel footnotes.


ekl176 (commented):
✅

PR #2370, opened 2026-08-12
RISDEV-10568 badges for meta data items

UrsKahmann:
**Metadata**
- refactor Metadata component to enable rendering values as badges

**Badge**
- add a new GRAY badge variant
- add new prop "variant" to expose different typography variants (will be needed for RISDEV-11103 RISDEV-12247)

**Notes**
- I am not super happy with the prop "variant": Not sure if the naming makes sense. Based on the values it should probably rather be called "size".. But also the typo variants don't really follow this strict size pattern (small, medium, large, etc...). So not sure if even better value names would be more descriptive, but I couldn't think of any. 
- I will remove the color enum and use a string union type for the colors instead in a separate PR

RISDEV-10568

### DIFF
--- a/backend/e2e-data/literature/XXLS000000001.akn.xml
+++ b/backend/e2e-data/literature/XXLS000000001.akn.xml
@@ -59,9 +59,6 @@
                     <akn:implicitReference showAs="FooBar, 1982, 123-123">
                         <ris:fundstelleUnselbstaendig periodikum="FooBar" zitatstelle="1982, 123-123"/>
                     </akn:implicitReference>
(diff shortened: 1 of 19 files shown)


andreasphil (commented):
great work! I agree the naming for the size variants is a bit awkward, but I don't have a better idea. Let's keep it for now. Suggestion: add a comment to each saying where it's meant to be used, that will make it easier for us to map the variant names to the design spec. (I don't think we should take the design spec names as they describe *where* it's meant to be used, not *what* it is)

PR #2377, opened 2026-08-13
RISDEV-10568 use string union type for badge colors

UrsKahmann:
RISDEV-10568


andreasphil (commented):
thanks, I never liked this ENUM 🎊

PR #2378, opened 2026-08-13
RISDEV-0000 consistent loading states

andreasphil:
This PR unifies all page-level data fetching to follow the same, consistent pattern:

- data loading happens at the beginning of the page and is `await`-ed, blocking any further rendering
- immediately after, results are checked for errors or missing required data; error handling is triggered
- cleans up templates and code paths that are unreachable now that pages can better rely on their data
- parallelizes some requests that used to run one after another, despite being independent
- consistently uses Nuxt's loading indicator + loading state on the search button to indicate activity on search pages
- adds some missing `await`s in E2E tests along the way that should mitigate flakyness

### DIFF
--- a/frontend/src/pages/translations/[id].vue
+++ b/frontend/src/pages/translations/[id].vue
@@ -28,12 +28,14 @@ useHead({
 const route = useRoute();
 const id = route.params.id as string;
 
-const { data, error } = await fetchTranslationAndHTML(id);
+const [{ data, error }, { legislation }] = await Promise.all([
+  fetchTranslationAndHTML(id),
+  getGermanOriginal(id),
+]);
+
 if (error.value || !data.value) {
   throw createError({ statusCode: error.value?.status ?? 500 });
 }
-
-const { legislation } = await getGermanOriginal(id);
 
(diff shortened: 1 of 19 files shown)


frontend/src/pages/translations/[id].vue
```diff
@@ -28,13 +28,15 @@ useHead({
 const route = useRoute();
 const id = route.params.id as string;
 
-const { data, error } = await fetchTranslationAndHTML(id);
+const [{ data, error }, { legislation }] = await Promise.all([
+  fetchTranslationAndHTML(id),
+  getGermanOriginal(id),
+]);
+
 if (error.value || !data.value) {
```
Comment (UrsKahmann) [inline frontend/src/pages/translations/[id].vue:36]: **question:** For the other documents we don't do the `|| !data.value` check.. We could probably remove it here as well?

UrsKahmann (commented):
Awesome! This is a nice improvement 🚀

frontend/src/pages/translations/[id].vue
Comment (andreasphil) [reply frontend/src/pages/translations/[id].vue:36]: fixed in #2386

PR #2380, opened 2026-08-14
RISDEV-11103 badges in details section

UrsKahmann:
- display norm references and fundstellen as badges in literature and VV  details list

RISDEV-11103


andreasphil (commented):
👍

PR #2381, opened 2026-08-14
RISDEV-0000 tidy Node and pnpm versions management

andreasphil:
- centralized Node and pnpm versions in the `devEngines` package property
- replaces the deprecated `pnpm/action-setup` with `pnpm/setup` in CI


UrsKahmann (approved):
👏

PR #2386, opened 2026-08-14
RISDEV-0000 various cleanup and simplifications in the frontend

andreasphil:
- removes an unnecessary `data` check, fixing PR review feedback from #2378 
- sends the `Accept` header for caselaw HTML, consistent with other HTML requests
- reduces some needlessly broad return values, reducing API surface
- removes redundant custom error handling from translations-related API composables
- prevent simple search from firing immediately, consistent with advanced search (page controls when the search should be submitted)
- replace `withDefaults` in pagination, simplify props access
- turns a value that never changes from a `computed` to a simple variable
- makes the court filter use the RIS backend API helper instead of plain fetch

### DIFF
--- a/frontend/src/composables/useTranslationData.ts
+++ b/frontend/src/composables/useTranslationData.ts
@@ -42,29 +42,17 @@ function legislationSearchURL(id: string, currentDate: string) {
 }
 
 export async function fetchTranslationList() {
-  const { data, error, status, pending, execute } = await useRisBackend<
-    TranslationContent[]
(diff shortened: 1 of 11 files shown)


frontend/src/composables/useTranslationData.ts
UrsKahmann, line 1: **praise:** Nice, thanks for the simplification! 🚀

UrsKahmann (commented):
✅

PR #2395, opened 2026-08-17
RISDEV-12247 badges in search results

UrsKahmann:
- remove icon and display document type in bold and first place in search result header
- move search result header validity status badge into the row with other items
- display fundstellen and aktenzeichen as badges in search result header
- display validity status of norms in metadata section as badges
- generally I tried to:
    - align the search result item creation across document types (only adding item if it has value, instead of adding all and then filtering out empty ones)
    - remove markup support in places where it is not needed (yet?). If needed in the future it can be added then. (e.g. only the badge item type in the search result header has markup support now, the simple text ones don't)

### Things I was unsure about
- does it make sense to have the `documentType` as extra property in the `SearchResultHeader` component? Should simply the first item in the list be the doc type rendered bold?

RISDEV-12247

### DIFF
--- a/frontend/src/assets/typography.css
+++ b/frontend/src/assets/typography.css
@@ -47,11 +47,11 @@
 }
 
 @utility typo-label2-regular {
-  @apply ris-label2-regular 2xl:text-[1.125rem];
+  @apply ris-label2-regular 2xl:ris-label1-regular;
 }
 
 @utility typo-label2-bold {
-  @apply ris-label2-bold 2xl:text-[1.125rem];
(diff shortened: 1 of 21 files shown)


andreasphil (commented):
👍

andreasphil:
> does it make sense to have the documentType as extra property in the SearchResultHeader component? Should simply the first item in the list be the doc type rendered bold?

I prefer to have the `documentType` prop over rendering whatever the first item is as bold. It's more readable 👍

PR #2397, opened 2026-08-17
RISDEV-12275, RISDEV-12144, RISDEV-12142 log level adjustments

andreasphil:
This PR adjusts some logging to reduce noise in Sentry:

- RISDEV-12144: Connection closed by peer should be a warning
- RISDEV-12275: `AbortException` is a normal part of how Nuxt operates, it should not be reported

It also changes logging of invalid user-submitted queries (RISDEV-12142), although that is a bit more involved than a simple logging level change:

- Controllers catch `UncategorizedElasticsearchException`s, but the retry template throws `RetryException`s that contain the original exception. The controller's `catch` therefore never matched. Added a helper that unwraps the `RetryException`, so the endpoint now correctly returns a 422 instead of 500 and logs the issue as a warning.
- Extended `LuceneQueryTools.checkForInvalidQuery` to also detect parsing errors similar to other error types.
- The 422 handler in `ControllerExceptionHandler` logged `exception.getMessage()`, which is always `null` on `CustomValidationException`. It now logs the errors.
- The retry policy retried things three times even if we know they won't succeed, so an unparseable query causes 3 warnings before erroring. Excluded client-error responses from retrying.

---

@andrew-cenkner-digitalservice: I went into this thinking it would be a simple log level change. It turned out to be a bit more complicated. Disclaimer: the 2 commits for RISDEV-12142 have been written by Claude, I think the explanation and code changes make sense but I don't 100% understand them, or all of the implications for the backend. Let me know if you think they're useless. Happy to drop those commits then.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
@@ -1,12 +1,15 @@
 package de.bund.digitalservice.ris.search.config.opensearch;
 
 import java.time.Duration;
+import lombok.SneakyThrows;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.retry.RetryException;
 import org.springframework.core.retry.RetryPolicy;
 import org.springframework.core.retry.RetryTemplate;
+import org.springframework.core.retry.Retryable;
 
 /** RetryConfiguration fo configure behaviour of opensearch query retries */
 @Configuration
@@ -36,5 +39,28 @@ public class OpensearchRetryConfiguration {
             .build();
 
     return new RetryTemplate(retryPolicy);
+  }
+
+  /**
+   * Runs an operation through the given retry template, unwrapping the {@link RetryException} that
+   * Spring throws once the retries are exhausted.
+   *
+   * <p>Without unwrapping, callers only see a generic retry error and can no longer tell what made
+   * the operation fail, e.g. that OpenSearch rejected the query as unparseable. Rethrowing the
+   * original failure keeps the exceptions callers handle the same as if the operation hadn't been
+   * retried at all.
+   *
+   * @param <T> the type the operation returns
+   * @param retryTemplate the retry template controlling the retry behaviour
+   * @param operation the operation to execute
+   * @return the result of the operation
+   */
+  @SneakyThrows
+  public static <T> T executeWithRetries(RetryTemplate retryTemplate, Retryable<T> operation) {
+    try {
+      return retryTemplate.execute(operation);
+    } catch (RetryException e) {
+      throw e.getCause();
+    }
   }
 }
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
@@ -1,12 +1,15 @@
 package de.bund.digitalservice.ris.search.config.opensearch;
 
 import java.time.Duration;
+import lombok.SneakyThrows;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.retry.RetryException;
 import org.springframework.core.retry.RetryPolicy;
 import org.springframework.core.retry.RetryTemplate;
+import org.springframework.core.retry.Retryable;
 import org.springframework.data.elasticsearch.UncategorizedElasticsearchException;
 import org.springframework.http.HttpStatusCode;
 
@@ -45,6 +48,30 @@ public class OpensearchRetryConfiguration {
             .build();
 
     return new RetryTemplate(retryPolicy);
+  }
+
+  /**
+   * Runs an operation through the given retry template, unwrapping the {@link RetryException} that
+   * Spring throws once the operation either exhausts its retries or fails with an error the {@link
+   * RetryPolicy} decided not to retry.
+   *
+   * <p>Without unwrapping, callers only see a generic retry error and can no longer tell what made
+   * the operation fail, e.g. that OpenSearch rejected the query as unparseable. Rethrowing the
+   * original failure keeps the exceptions callers handle the same as if the operation hadn't been
+   * retried at all.
+   *
+   * @param <T> the type the operation returns
+   * @param retryTemplate the retry template controlling the retry behaviour
+   * @param operation the operation to execute
+   * @return the result of the operation
+   */
+  @SneakyThrows
+  public static <T> T executeWithRetries(RetryTemplate retryTemplate, Retryable<T> operation) {
+    try {
+      return retryTemplate.execute(operation);
+    } catch (RetryException e) {
+      throw e.getCause();
+    }
   }
 
   /**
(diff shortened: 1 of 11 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/RestClientConfig.java
```diff
@@ -93,10 +92,10 @@ public ElasticsearchOperations elasticsearchOperations(
...
 
-      @SneakyThrows
       @Override
       public <T> T execute(OpenSearchRestTemplate.ClientCallback<T> callback) {
-        return retryTemplate.execute(() -> super.execute(callback));
+        return OpensearchRetryConfiguration.executeWithRetries(
+            retryTemplate, () -> super.execute(callback));
       }
     };
   }
```
andrew-cenkner-digitalservice, line 100: Since this method is exactly the same in RestClientConfig and RestClientConfigDev, I would extract it to AbstractOpenSearchConfiguration.

Normally that would not be in the scope of this ticket, but it enables you to delete the static method
```
  @SneakyThrows
  public static <T> T executeWithRetries(RetryTemplate retryTemplate, Retryable<T> operation) {
    try {
      return retryTemplate.execute(operation);
    } catch (RetryException e) {
      throw e.getCause();
    }
```
and instead put the logic directly inside the `execute` method.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/ControllerExceptionHandler.java
```diff
@@ -245,7 +251,12 @@ public ResponseEntity<CustomErrorResponse> handleRouteNotFoundException(
   @ExceptionHandler(OpenSearchFetchException.class)
   public ResponseEntity<CustomErrorResponse> handleElasticsearchException(
       OpenSearchFetchException ex) {
-    logger.error("Opensearch fetch error", ex);
+    if (NestedExceptionUtils.getMostSpecificCause(ex) instanceof ConnectionClosedException) {
```
andrew-cenkner-digitalservice: Does ConnectionClosedException really only represent connections closed by the client?

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/ControllerExceptionHandler.java
```diff
@@ -237,6 +238,11 @@ public ResponseEntity<CustomErrorResponse> handleRouteNotFoundException(
    * Handles exceptions of type {@link OpenSearchFetchException} and returns a standardized error
    * response with HTTP status 500 (Internal Server Error).
    *
+   * <p>A connection from the client's pool that OpenSearch has closed on its side surfaces as a
```
andrew-cenkner-digitalservice, line 242: I'm assuming this documentation is explaining that it's safe to use warn when there was a client error. In particular connection closed by client. Can you please change the documentation to make this more clear?

backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
andrew-cenkner-digitalservice, line 46: I think we want 

- First client failure -> Warning and no retries
- Last server failure -> Error and no more retries
- Non last server failure -> Warning and retry

**EDIT : Please read the comment about `@ExceptionHandler` below before trying out a RetryListener. It might be possible that the last server failure case doesn't need to be handled here. Then I would just change so both client and server log a warning here.**

Unfortunately it seems spring is a bit rigid here and requires more boilerplate. Could you please look into something like
```
  @Bean
  public RetryTemplate openSearchRetryTemplate() {
      RetryPolicy retryPolicy =
              RetryPolicy.builder()
                      .maxRetries(2)
                      .delay(Duration.ofSeconds(1))
                      .multiplier(2.0)
                      .predicate(throwable -> {
                          if (true) {
                              logger.warn( "< client failure warning here >");
                              return false;
                          } else {
                              logger.warn("<non last server failure warning here>");
                              return true;
                          }
                      })

              .build();
     RetryTemplate t = new RetryTemplate(retryPolicy);
      t.setRetryListener(new RetryListener() {
          @Override
          public void onRetryPolicyExhaustion(RetryPolicy retryPolicy, Retryable<?> retryable, RetryException exception) {
              logTerminalServerFailure();
              }

          @Override
          public void onRetryPolicyInterruption(RetryPolicy retryPolicy, Retryable<?> retryable, RetryException exception) {
              logTerminalServerFailure();
          }
          @Override
          public void onRetryPolicyTimeout(RetryPolicy retryPolicy, Retryable<?> retryable, RetryException exception) {
              logTerminalServerFailure();
          }
              private void logTerminalServerFailure(){
                  logger.error( "< server failure error here >");
              }
      });
    return t;
  }
```

backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
```diff
@@ -37,4 +45,44 @@ public RetryTemplate openSearchRetryTemplate() {
...
+   * @param retryTemplate the retry template controlling the retry behaviour
+   * @param operation the operation to execute
+   * @return the result of the operation
+   */
+  @SneakyThrows
+  public static <T> T executeWithRetries(RetryTemplate retryTemplate, Retryable<T> operation) {
+    try {
+      return retryTemplate.execute(operation);
+    } catch (RetryException e) {
+      throw e.getCause();
```
andrew-cenkner-digitalservice, line 78: I think the consequence of this is that previously UncategorizedElasticsearchException (and possibly other exceptions) were silently dropped and now they will bubble up to spring top level.
Normally we would have a `@ExceptionHandler` in ControllerExceptionHandler, but I don't see one for UncategorizedElasticsearchException and I also don't see a test case for when UncategorizedElasticsearchException happens 3 times.
I'm guessing that means it will be caught by
```
  @ExceptionHandler({
    IllegalArgumentException.class,
    FileTransformationException.class,
    Exception.class
  })
```
Maybe if we make the proper `@ExceptionHandler` we could log a clear error there and avoid the `RetryListener` I mentioned in the other comment.

andrew-cenkner-digitalservice (changes requested):
Please see individual comments.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/ControllerExceptionHandler.java
andreasphil: Proposed fix after discussing:

- Also check that message is "closed by peer" → then log as warning
- Otherwise log as error

backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/OpensearchRetryConfiguration.java
andreasphil: Proposed fix:

- Add explicit exception handler for `UncategorizedElasticsearchException`
- It's ok if other, so far unhandled exception, bubble up and get caught by the general handler

To test this:

- integration tests + hook into one of the beans that already exist
- throw an exception in the mock response

Simplified example:

```java
@SpringBootTest
class MyServiceTest {

    @Autowired
    private MyService myService; // The real bean under test

    @MockitoBean
    private DependencyService dependencyService; // This replaces the real bean with a mock

    @Test
    void testMockedBehavior() {
        // 1. Setup the mock behavior
        Mockito.when(dependencyService.getData()).thenReturn("Mocked Data");

        // 2. Call the real service, which uses the mock internally
        String result = myService.doSomething();

        // 3. Assert
        Assertions.assertEquals("Expected Result", result);
    }
}
```

backend/src/main/java/de/bund/digitalservice/ris/search/config/opensearch/RestClientConfig.java
andreasphil: As discussed: put the logic directly into `elasticsearchOperations` of both REST client configs.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/ControllerExceptionHandler.java
andreasphil: Obsolete now

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/ControllerExceptionHandler.java
andreasphil: Fixed

andrew-cenkner-digitalservice (approved):

PR #2398, opened 2026-08-18
RISDEV-12193 migrate panel menu component

andreasphil:
This PR replaces the panel menu component. One of the problems of the panel menu was that it wasn't really intended to be used in the way we used it, leading to a number of accessibility problems (see e.g. RISDEV-10228). The new version therefore tries a different approach: since we use the panel menu as a filter selection, the new component is implemented as radio buttons under the hood:

- combined list and radio semantics are better accessibility for this context
- we get keyboard control via up/down arrows for free
- open/close can be simplified with nested lists and children showing when the parent is selected (nesting can only be 1 level deep)

3 implications:

- when selecting "Gerichtsentscheidungen", we want to show the different types alongside "alle Gerichtsentscheidungen", which is identical to "Gerichtsentscheidungen". To support this, a parent item can have an alias (`self...` in the component) that will be rendered as its first child. When selecting the parent, the value and focus will automatically move to the alias child.
- the panel menu required pressing <kbd>Enter</kbd> when navigating to a document kind via keyboard before applying the selection. The radio implementation doesn't have this intermediary step, it updates the filter and submits the search immediately. For mouse users, the behavior is identical. Keyboard users may fire a few unnecessary searches while navigating to the document kind they want. I think this is ok because
  - most value changes on the search pages are instant anyway
  - Nuxt cancels obsolete requests, so our frontend will never get confused
  - it only applies to keyboard use
- showing the caselaw subtypes only when selecting "Gerichtsentscheidungen" means that the screen reader goes from something like "list, 5 items" to "list, 8 items". I don't know if this is confusing to some people, but I think it makes sense because it mirrors what you see, and the new announcements actually make it easier to navigate than the previous component. Furthermore, when navigating with the keyboard from "Verwaltungsvorschriften" up, you first reach "Alle Gerichtsentscheidungen". After that you can press down again to get to the subtypes. Again—potentially confusing, but also mirrors exactly the behavior that mouse users get, so it's logical in my opinion.

---

this is an experiment for now, pending design review. Might revert the commit again later


UrsKahmann (commented):
Great work!!! 📻 🔘

PR #2402, opened 2026-08-18
RISDEV-12145 Find cause of xslt transformation error

andrew-cenkner-digitalservice:
*Refactor encoding logic to the controller.
*Added a validation to only call xslt if the article actually exists
*Changed a test case

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/NormsController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/NormsController.java
@@ -22,6 +22,7 @@ import de.bund.digitalservice.ris.search.schema.CollectionSchema;
 import de.bund.digitalservice.ris.search.schema.LegislationExpressionSchema;
 import de.bund.digitalservice.ris.search.schema.LegislationExpressionSearchSchema;
 import de.bund.digitalservice.ris.search.schema.SearchMemberSchema;
+import de.bund.digitalservice.ris.search.service.ArticleService;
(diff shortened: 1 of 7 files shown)


andreasphil (approved):

andreasphil:
Backend changes look good, judging from the failing E2E tests it seems the frontend needs some adjustments too?

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/NormsController.java
```diff
@@ -553,27 +560,53 @@ public ResponseEntity<String> getLegislationArticleAsHtml(
...
-            "xml");
-    final Optional<byte[]> normFileByEli = normsService.getNormFileByEli(eli);
-    return normFileByEli
-        .map(
-            bytes ->
-                ResponseEntity.ok(
-                    xsltTransformerService.transformArticle(bytes, articleEid, resourceBasePath)))
-        .orElseGet(() -> ResponseEntity.notFound().build());
+  private Optional<String> getActualEid(String expressionEli, String eidGiven) {
+    // An eid can contain % and therefore can't be directly used as a path variable. When it does
```
ekl176: This might be a public method on the ArticleService instead.

andrew-cenkner-digitalservice: Makes sense. I will move it there. I also want to discuss the approach to eid in general with the other engineers because this fallback encoding is not great.

PR #2405, opened 2026-08-19
RISDEV-12193 ensure full size clickable area on iOS

andreasphil:
In iOS Safari, absolute positioning with `inset-0` is not enough to make the input span the entire control, resulting in a tiny clickable area. Additionally setting explicit width and height should fix it.


UrsKahmann (commented):
✅

PR #2462, opened 2026-08-31
Format previousDecisions also

MagdaN:


backend/src/test/java/de/bund/digitalservice/ris/search/integration/controller/api/RechtsprechungControllerTest.java
```diff
@@ -126,9 +126,15 @@ void shouldReturnSingleCaselawJson() throws Exception {
...
+                .value(
+                    "previous decision court type previous decision court location, previous"
+                        + " decision document type vom 1. Januar 2020 - previous decision file"
+                        + " number"),
             jsonPath("$.vorgehendeEntscheidungen[1]")
-                .value("previous decision file number, previous decision court type"),
+                .value(
+                    "previous decision court type previous decision court location, previous"
+                        + " decision document type vom 1. Januar 2020 - previous decision file"
+                        + " number"),
```
ekl176: I'd rather have different content between the fields of 2 items of the same list. By checking both entries of the vorgehendeEntscheidungen list with the same content, it is more likely to create a false positive in the test cases.

PR #2465, opened 2026-08-31
Risdev-11410 legislation jsonld context

ekl176:
This PR adds an Endpoint that serves a jsonld schema in version 1.1 to be used across all our endpoints. 
The absolute path of the new endpoint is to be used as the remote context in our api responses. 

The context only contains the properties used in the NormController right now. The context will be extended in subsequent PRs.

Legislation schemas and the CollectionSchema were given a @context field that points to the remote context. 

A ResultMatcher is added to check if every Field in an API response is extended. It can not test if the fields are semantically correct.


backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/JsonLdContextController.java
```diff
@@ -0,0 +1,23 @@
...
+/** Controller to serve the jsonLd context */
+@RestController
+public class JsonLdContextController {
+
+  private final Resource contextResource = new ClassPathResource("/jsonld/1_1/v1.jsonld");
+
+  @GetMapping(path = ApiConfig.Paths.JSONLD_CONTEXT, produces = "application/ld+json")
+  public ResponseEntity<Resource> getContext() {
+    return ResponseEntity.ok()
+        .header(HttpHeaders.CACHE_CONTROL, "public, max-age=31536000, immutable")
```
andrew-cenkner-digitalservice: This would mean we only update the jsonld context when the major api version (currently "v1") changes. We should either add sub versioning on this file, or lower the time span of 1 year.

backend/src/main/resources/application-dev.yaml
```diff
@@ -1,5 +1,6 @@
 server:
   front-end-url: "http://localhost:3000/"
+  back-end-url: "http://localhost:8080/"
```
andrew-cenkner-digitalservice: I would remove the trailing / here so we could clean up the usage (see related comment in JsonLdUtils.java)

backend/src/main/java/de/bund/digitalservice/ris/search/utils/JsonLdUtils.java
```diff
@@ -1,35 +1,22 @@
...
-/**
- * A utility class for working with JSON-LD (JSON for Linked Data). This class provides methods for
- * serializing objects into JSON-LD format using a customized instance of the Jackson {@link
- * ObjectMapper}.
- *
- * <p>This class is not instantiable and cannot be extended.
- */
+/** A utility class for working with JSON-LD (JSON for Linked Data). */
+@Component
 public class JsonLdUtils {
```
andrew-cenkner-digitalservice: I would delete this class. It's quite heavy for what it actually does. If we remove the trailing / from the properties then it's essentially just doing
```
${server.back-end-url} + ApiConfig.Paths.JSONLD_CONTEXT
```
which could be inlined in the 4 places that it is used. See comment in NormsController.java for example suggestion.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/NormsController.java
```diff
@@ -97,11 +99,13 @@ public NormsController(
       NormsService normsService,
       ArticleService articleService,
       NormXsltTransformerService xsltTransformerService,
-      ChangelogService<NormsBucket> changelogService) {
+      ChangelogService<NormsBucket> changelogService,
+      JsonLdUtils jsonLdUtils) {
```
andrew-cenkner-digitalservice: You can instead add
```
      @Value("${server.back-end-url}") String backEndUrl
```
here and use
```
backEndUrl + ApiConfig.Paths.JSONLD_CONTEXT
```
later instead of
```
jsonLdUtils.getJsonldPath()
```

andrew-cenkner-digitalservice (changes requested):

backend/src/main/resources/application-dev.yaml
ekl176: Just kept the trailing / for the sake of consistency to the other configs. As discussed, I'll remove it from the back-end-url and we'll remove the trailing slashes in the configs as well in later PRs.

PR #2466, opened 2026-08-31
Add missing cl meta data to details tab

MagdaN:
This PR adds the case law fields that were already exposed on the caselaw API endpoint — gesetzeskraft, streitjahre, vorgehendeEntscheidungen, nachgehendeEntscheidungen — to the case law detail page's UI:

- Details tab: renders "Gesetzeskraft", "Streitjahre", "Vorgehende Entscheidungen", and "Nachgehende Entscheidungen" as new list entries.
- DetailsList.vue: fixed a layout bug where a long label wrapping to two lines left a visual gap before the next value, and added a labelClass prop so specific items (the two "…Entscheidungen" labels) can force a two-line break via whitespace-pre-line regardless of available width.
- E2E fixture (BDRE000800001.xml): extended with multiple previous/ensuing decisions and a gesetzeskraft/streitjahre example, to exercise the new fields.
- E2E test: added coverage asserting the full formatted details list for BDRE000800001, including all four new fields.

### DIFF
--- a/frontend/src/pages/gerichtsentscheidungen/[documentNumber].vue
+++ b/frontend/src/pages/gerichtsentscheidungen/[documentNumber].vue
@@ -118,6 +118,29 @@ const detailItems = computed<DetailsListItem[]>(() => [
     value: formatArray(caseLaw.value?.decisionName ?? []),
   },
   {
+    type: "list",
+    label: "Gesetzeskraft:",
+    values: caseLaw.value?.gesetzeskraft ?? [],
+  },
+  {
+    type: "list",
+    label: "Streitjahre:",
+    values: caseLaw.value?.streitjahre ?? [],
+  },
+  {
+    type: "list",
+    label: "Vorgehende\nEntscheidungen:",
+    labelClass: "md:whitespace-pre-line",
+    values: caseLaw.value?.previousDecisions ?? [],
+  },
+  {
+    type: "list",
+    label: "Nachgehende\nEntscheidungen:",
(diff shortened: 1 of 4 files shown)


frontend/src/pages/gerichtsentscheidungen/[documentNumber].vue
```diff
@@ -117,6 +117,29 @@ const detailItems = computed<DetailsListItem[]>(() => [
...
+  },
+  {
+    type: "list",
+    label: "Streitjahre:",
+    values: caseLaw.value?.streitjahre ?? [],
+  },
+  {
+    type: "list",
+    label: "Vorgehende\nEntscheidungen:",
+    labelClass: "md:whitespace-pre-line",
```
Comment (UrsKahmann) [inline frontend/src/pages/gerichtsentscheidungen/[documentNumber].vue:133]: **todo:** Remove the newline characters from the labels (layout should not be mixed with the content) and also adding the `labelClass` prop just to style these two entries differently feels wrong.  Ideally the `dt` element in the details list should have a styling that works for labels of different lengths so the layout is consistent and there is now need for custom styling certain labels here. 

From quickly looking at it removing the newlines and the labelClass and then adding a `text-balance` to the `dt` element class results in the same effect. This still leaves the problem that for multiline lables with single line values the `h-0` causes the gap to collapse -> there is no space to the next item row.. But I couldn't figure out a quick solution for that. I think the combination of grid + description list might not allow a clean solution for that 🤔

UrsKahmann (commented):
Looks good in general, I left one comment regarding the DetailsList styling :)

PR #2477, opened 2026-09-03
Risdev 12358 rework targeted article search

andrew-cenkner-digitalservice:
* Refactored the Article class to use a static Fields class like the other entities improving clarity and reducing duplication of static constants
* Refactored search_keyword to article_fingerprint which is more descriptive and also uses the fingerprint analyzer instead of slop
* Introduced a new method buildID to remove duplicated code when we need to build an article id in different places.
* Added article_fingerprints to the norms index in the same pattern as article_texts
* Added a targeted search on article_fingerprints so when there is a match it impacts the main ranking and not just the sub ranking


UrsKahmann (changes requested):
Looks good, the only todo I see is: Tests 😉
I see most of this is not easy to test, but this adds a new field to the Norm entity -> the `NormLdmlToOpenSearchMapperTest` should cover that the fingerprint is mapped as expected. 
Also the boost that is added for the article name + abbreviation combination should probably be tested as well, although I see that this might not be that easy.

andrew-cenkner-digitalservice:
> Looks good, the only todo I see is: Tests 😉 I see most of this is not easy to test, but this adds a new field to the Norm entity -> the `NormLdmlToOpenSearchMapperTest` should cover that the fingerprint is mapped as expected. Also the boost that is added for the article name + abbreviation combination should probably be tested as well, although I see that this might not be that easy.

It's not very clear, but the mapping is already tested in tests like AdvancedSearchControllerApiTest::shouldReturnOkArticleNamesQuery .

We don't have a good way to test ranking yet, but I started building a way to test ranking in a different PR for case law.

UrsKahmann:
> > Looks good, the only todo I see is: Tests 😉 I see most of this is not easy to test, but this adds a new field to the Norm entity -> the `NormLdmlToOpenSearchMapperTest` should cover that the fingerprint is mapped as expected. Also the boost that is added for the article name + abbreviation combination should probably be tested as well, although I see that this might not be that easy.
> 
> It's not very clear, but the mapping is already tested in tests like AdvancedSearchControllerApiTest::shouldReturnOkArticleNamesQuery .
> 
> We don't have a good way to test ranking yet, but I started building a way to test ranking in a different PR for case law.

Not really, no test covers that the `NormLdmlToOpenSearchMapper` actually sets the `articleFingerprints` field correctly. In the mapper, if you replace `.articleFingerprints(articleFingerprints)` with `.articleFingerprints(List.of("foo"))` all tests (unit and integration tests) pass...

andrew-cenkner-digitalservice:
> > > Looks good, the only todo I see is: Tests 😉 I see most of this is not easy to test, but this adds a new field to the Norm entity -> the `NormLdmlToOpenSearchMapperTest` should cover that the fingerprint is mapped as expected. Also the boost that is added for the article name + abbreviation combination should probably be tested as well, although I see that this might not be that easy.
> > 
> > 
> > It's not very clear, but the mapping is already tested in tests like AdvancedSearchControllerApiTest::shouldReturnOkArticleNamesQuery .
> > We don't have a good way to test ranking yet, but I started building a way to test ranking in a different PR for case law.
> 
> Not really, no test covers that the `NormLdmlToOpenSearchMapper` actually sets the `articleFingerprints` field correctly. In the mapper, if you replace `.articleFingerprints(articleFingerprints)` with `.articleFingerprints(List.of("foo"))` all tests (unit and integration tests) pass...

Ahh, I think I understand the confusion now. There used to be a magic number `3` in the code. This implies there was some kind of special mapping where the count and order of tokens was important. This was previously a bug and part of the motivation to switch to fingerprint. Searching with "§ 75 fictional law" and "§ 75 law fictional" were valid "targeted article search" and the magic number of 3 won't work because there are 4 tokens. It just happened to mostly work in the past.

The opensearch field mapping actually stayed the same (stayed `text`) and therefore there should not be any change to mapping tests. `foo` technically was and still is valid according to the mapping, even though single token values will never happen in practice.

What did change was the analyzer, which will impact the filtering and ranking logic. I plan to write tests for the "§ 75 fictional law" case above where it should pass with the changes and would have previously failed, but am waiting to merge this PR https://github.com/digitalservicebund/ris-search/pull/2478 which has general test framework changes to make writing ranking tests easier.

andrew-cenkner-digitalservice:
> > > Looks good, the only todo I see is: Tests 😉 I see most of this is not easy to test, but this adds a new field to the Norm entity -> the `NormLdmlToOpenSearchMapperTest` should cover that the fingerprint is mapped as expected. Also the boost that is added for the article name + abbreviation combination should probably be tested as well, although I see that this might not be that easy.
> > 
> > 
> > It's not very clear, but the mapping is already tested in tests like AdvancedSearchControllerApiTest::shouldReturnOkArticleNamesQuery .
> > We don't have a good way to test ranking yet, but I started building a way to test ranking in a different PR for case law.
> 
> Not really, no test covers that the `NormLdmlToOpenSearchMapper` actually sets the `articleFingerprints` field correctly. In the mapper, if you replace `.articleFingerprints(articleFingerprints)` with `.articleFingerprints(List.of("foo"))` all tests (unit and integration tests) pass...

I added a test in NormLdmlToOpenSearchMapperTest

PR #2479, opened 2026-09-03
RISDEV-11495 change court filter label and add description

UrsKahmann:
- change court filter label from "Bundesgericht" to "Gericht"
- add a description

RISDEV-11495

**Info for Reviewer**
I didn't change that much in the unit test, the indentation was updated, thats why it looks like a lot of changed, sorryy!


UrsKahmann:
This implemetation broke the crout dropdown visually. As there is no easy way with the primevue combobox to attach the text as description to the input, this has been fixed here https://github.com/digitalservicebund/ris-search/pull/2481 without an id link. This can be revisited when the primevue component is replaced.

MagdaN (commented):
Looks good, thank you!

PR #2481, opened 2026-09-03
RISDEV-11495 fix court dropdown

UrsKahmann:
The previous implementation broke the combobox visually. This is a quick fix.

RISDEV-11495


MagdaN (commented):
Thank you, looks good!

PR #2484, opened 2026-09-04
Mirror caselaw api andpoints on new path

MagdaN:


backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/RechtsprechungSearchController.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Controller class for handling search of Rechtsprechung (case law) documents. */
+@Tag(name = "Rechtsprechung")
+@RestController
+@Profile({"dev", "e2e", "default", "staging", "uat", "test", "prototype"})
+public class RechtsprechungSearchController {
```
ekl176: When we create the new Rechtsprechung Controller from scratch we might want to use it as an opportunity to put all functionality in one controller instead of differentiating between searchController.

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/RechtsprechungSearchController.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.data.elasticsearch.core.SearchPage;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Controller class for handling search of Rechtsprechung (case law) documents. */
+@Tag(name = "Rechtsprechung")
+@RestController
+@Profile({"dev", "e2e", "default", "staging", "uat", "test", "prototype"})
```
ekl176: is there a specific reason why the new controller shouldn't be in production? If not we can remove the whole profile here

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/RechtsprechungSearchController.java
```diff
@@ -0,0 +1,121 @@
...
+   */
+  @GetMapping(path = ApiConfig.Paths.RECHTSPRECHUNG, produces = MediaType.APPLICATION_JSON_VALUE)
+  @Operation(
+      operationId = "searchRechtsprechung",
+      summary = "List and search decisions",
+      description =
+          "The endpoint returns a list of decisions from our database. The list is paginated and can be filtered and sorted.")
+  @ApiResponse(responseCode = "200", description = "Success")
+  @ApiResponse(responseCode = "500", description = "Internal Server Error", content = @Content)
+  public ResponseEntity<CollectionSchema<SearchMemberSchema<CaseLawSearchSchema>>> searchAndFilter(
```
ekl176: The rechtsprechungController will still return english field names on its search endpoints.

PR #2497, opened 2026-09-07
RISDEV-12559 add administrative context

ekl176:
Add AdministrativeDirective context in jsonld. 
Use AdministrativeDirectiveSearchSchema in search Pages.

### DIFF
--- a/backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/AdministrativeDirectiveController.java
+++ b/backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/AdministrativeDirectiveController.java
@@ -3,6 +3,7 @@ package de.bund.digitalservice.ris.search.controller.api;
 import static org.springframework.http.HttpHeaders.CONTENT_DISPOSITION;
 
 import de.bund.digitalservice.ris.search.config.ApiConfig;
+import de.bund.digitalservice.ris.search.config.ServerConfig;
 import de.bund.digitalservice.ris.search.exception.CustomValidationException;
 import de.bund.digitalservice.ris.search.mapper.AdministrativeDirectiveSchemaMapper;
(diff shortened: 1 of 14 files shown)


backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/AdministrativeDirectiveController.java
```diff
@@ -106,7 +111,9 @@ public ResponseEntity<AdministrativeDirectiveSchema> getAdministrativeDirectiveM
     AdministrativeDirective unit = result.getFirst();
     return ResponseEntity.ok()
         .contentType(MediaType.APPLICATION_JSON)
-        .body(AdministrativeDirectiveSchemaMapper.fromDomain(unit));
+        .body(
+            AdministrativeDirectiveSchemaMapper.fromDomain(
+                unit, serverConfig.getBackEndUrl() + ApiConfig.Paths.JSONLD_CONTEXT));
```
UrsKahmann, line 116: **suggestion:** the construction of the context url is done in many places in the exact same way.

Idea: Instead of passing the full url to each mapper, only pass in the backend url. Then create a small helper method somewhere that takes the backend url and returns the json ld context url -> use that helper method in each mapper. 

This way if the construction of the full url ever needs to be changed it can be done in one place easily.

UrsKahmann (approved):

backend/src/main/java/de/bund/digitalservice/ris/search/controller/api/AdministrativeDirectiveController.java
ekl176: If the jsonld_context would change, we would only need to change the definition of the ApiConfig.Paths.JSONLD_CONTEXT. 

We can add a public static String jsonContext(String backendUrl) that appends the path though to keep it centralized.

PR #2500, opened 2026-09-08
Use new rechtsprechungs enpoints where possible

MagdaN:


frontend/src/composables/useCaselawSeo.ts
```diff
@@ -1,7 +1,7 @@
-import { type CaseLaw } from "~/types/api";
+import { type Rechtsprechung } from "~/types/api";
 
 export type UseCaselawSeoInput = {
-  caseLaw?: CaseLaw;
+  caseLaw?: Rechtsprechung;
```
UrsKahmann, line 4: **question:** To stay consistent, shouldn't we also rename the property to `rechtsprechung`?

frontend/src/composables/useCaselawSeo.ts
```diff
@@ -13,18 +13,16 @@ export function useCaselawSeo({ caseLaw, document }: UseCaselawSeoInput) {
   });
 }
 
-function buildTitle(caseLaw?: CaseLaw) {
-  return caseLaw?.headline || "Gerichtsentscheidung";
+function buildTitle(caseLaw?: Rechtsprechung) {
```
UrsKahmann, line 16: **suggestion** same here, rename the parameter to `rechtsprechung` as well?

frontend/src/composables/useCaselawSeo.ts
```diff
@@ -13,18 +13,16 @@ export function useCaselawSeo({ caseLaw, document }: UseCaselawSeoInput) {
...
+  return caseLaw?.kurztitel || "Gerichtsentscheidung";
 }
 
 function buildDescription(
-  caseLaw: CaseLaw | undefined,
+  caseLaw: Rechtsprechung | undefined,
```
UrsKahmann, line 21: same here

frontend/src/composables/useCaselawSeo.ts
```diff
@@ -43,16 +41,14 @@ function buildDescription(
 // TODO: The truncation can cause parts of the fileNumber to not be displayed
 // which is probably not the wanted behavior - this should be clarified and fixed in
 // https://digitalservicebund.atlassian.net/browse/RISDEV-11649
-function buildOgTitle(caseLaw?: CaseLaw) {
+function buildOgTitle(caseLaw?: Rechtsprechung) {
```
UrsKahmann, line 44: **suggestion:** Here too? Now I am also not so sure anymore..

frontend/src/types/api.ts
```diff
@@ -26,6 +26,7 @@ export type Statistics = components["schemas"]["StatisticsApiSchema"];
 
 // Case law
 export type CaseLaw = components["schemas"]["CaseLawSchema"];
```
UrsKahmann, line 28: **question:** Is this now even needed anymore?

UrsKahmann (approved):
Looks good, but this raises the more general question if/where we set the boundary of naming properties/parameters/variables german as well? 
I started commenting in one file but then wasn't so sure anymore. What I mean are things like this:
```
getCaselawSecondaryTitle(
    {
      decisionNames: caseLaw.value?.entscheidungsnamen ?? [],
      titleLine: caseLaw.value?.titelzeile,
    },
    false,
  ),
```
Where we now have a mix of english and german and need to keep in mind that "decisionNames" and "entscheidundsname" are the same from the domain perspective. But this is probably a more general topic to be discussed with the team and out of scope for this PR 🤔

PR #2502, opened 2026-09-09
RISDEV-11697 lucene query unexpected error

andrew-cenkner-digitalservice:
* Added IllegalArgumentException to the list of client errors to cause the API to return a client error and log at the WARN level (instead of ERROR level)


UrsKahmann:
✅

PR #2510, opened 2026-09-10
RISDEV-0000 split tests into dedicated source sets

UrsKahmann:
- Split backend tests into real test/integrationTest/dataTest/testFixtures Gradle source sets
- Moved tests with no 1:1 main-class match into a miscellaneous package per suite
- Fixed drifted test names (8 renames) and suffixed all integrationTest classes with IntegrationTest
- Cleaned up build.gradle.kts wiring
- the testFixtures source set is for utilties and resources that are shared across unit/integration/data test
- I actually don't really know how to run the data tests so I couldn't test if they still work as expected

RISDEV-0000


backend/build.gradle.kts
```diff
@@ -108,29 +110,87 @@ dependencies {
...
+val commonTestDependencies: JvmComponentDependencies.() -> Unit = {
+    implementation(libs.spring.boot.starter.test)
+    implementation(libs.spring.security.test)
+    implementation(libs.spring.boot.starter.webmvc.test)
+    implementation(libs.mockito.junit.jupiter)
+}
+
+testing {
+    suites {
+        val test by getting(JvmTestSuite::class) {
```
ekl176: JVMTestSuite seems to be marked unstable

backend/build.gradle.kts
```diff
@@ -108,29 +110,87 @@ dependencies {
...
+    testFixturesImplementation(libs.commons.io)
 }
 
 dependencyLocking {
     lockAllConfigurations()
 }
 
+// Test deps shared by the unit ("test") and "integrationTest" suites, which both boot a
+// Spring (Boot) test context.
+val commonTestDependencies: JvmComponentDependencies.() -> Unit = {
```
ekl176: JvmComponentDependencies seems to be marked unstable

backend/build.gradle.kts
```diff
@@ -108,29 +110,87 @@ dependencies {
...
+            useJUnitJupiter()
+            dependencies {
+                implementation(sourceSets["main"].output)
+                implementation(testFixtures(project()))
+                commonTestDependencies()
+                implementation(libs.archunit.junit5)
+            }
+        }
+
+        val integrationTest by registering(JvmTestSuite::class) {
```
ekl176: registering is marked as deprecated

backend/build.gradle.kts
UrsKahmann: Thanks, i replaced the other instructions that were deprecated/unstable but for this one i don't see an alternative 🤔

ekl176 (approved):

backend/build.gradle.kts
UrsKahmann: I refactored it but as dicsussed the whole jvm test suites plugin is marked as incubating so the warnigns probably won't go away. But the plugin exists for quite a while and [gradle recommends to use it](https://blog.gradle.org/introducing-test-suites#:~:text=Thus%2C%20Test%20Suites%20was%20born.%20The%20JVM,Test%20Suites%20in%20your%20JVM%20projects.%20Congratulations!) so I think we should be fine.

PR #2513, opened 2026-09-11
Fix adr numbering

malte-laukoetter:


UrsKahmann (approved):
Thanks for cleaning up 🧹 🙏

PR #2523, opened 2026-09-14
Risdev-12625 index article document number

ekl176:
Index document number as immutable identifier of an article across legislation expressions


backend/src/test/java/de/bund/digitalservice/ris/search/mapper/NormLdmlToOpenSearchMapperTest.java
```diff
@@ -577,4 +586,54 @@ private static Stream<Arguments> provideShortTitlePermutations() {
...
+                "art-z2",
+                null,
+                article -> {
+                  article
+                      .addHeading("Heading 2", null)
+                      .addParagraph(
+                          "Ein weiterer Satz mit einem Punkt in einer Aufzählung und noch einem Punkt.",
+                          "(1)")
+                      .addParagraph("Noch ein wichtiger Satz. Das ist der letzte Satz.", "(2)");
+                });
```
UrsKahmann: **suggestion:** As for this test the article content (e.g. heading, paragraph etc) don't matter, you could extend the `NormTestDataBuilder` by adding a function like:
```java
  /**
   * Adds an article to the norm's body and populates its body with default values.
   *
   * @param num the article number, e.g. "§ 1"
   * @param startDate date the article starts being valid
   * @param endDate date the article stops being valid
   * @param eId eId of the article, e.g. "art-z1"
   * @param dokNr document number mapped to this article
   * @return this builder for chaining
   */
  public NormTestDataBuilder article(
      String num, String startDate, String endDate, String eId, String dokNr) {
    this.article(
        num,
        startDate,
        endDate,
        eId,
        dokNr,
        article -> {
          article.addHeading("Heading", null).addParagraph("Some paragraph content", "(1)");
        });
    return this;
  }
```

And then this test could omit the article consumer in the data creation and would be a bit shorter and more focused? Just an idea

backend/src/testFixtures/java/de/bund/digitalservice/ris/builder/models/meta/proprietary/ris/RisDokNr.java
```diff
@@ -0,0 +1,19 @@
...
+import jakarta.xml.bind.annotation.XmlAttribute;
+import jakarta.xml.bind.annotation.XmlValue;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+/** Represents {@code ris:dokNr} mapping an eId to the immutable dokNr */
+@Getter
+@Setter
```
UrsKahmann, line 12: **todo:** Remove getter and setter annotation, they are not needed.

backend/src/testFixtures/java/de/bund/digitalservice/ris/builder/NormTestDataBuilder.java
```diff
@@ -385,14 +387,29 @@ public NormTestDataBuilder section(Section section) {
...
     this.document.getAct().getBody().addChild(article);
+
+    if (StringUtils.isNotEmpty(dokNr)) {
+      this.document
+          .getAct()
+          .getMeta()
+          .getProprietary()
+          .getRisMetadata()
+          .addDokNr(new RisDokNr("#" + eId, dokNr));
+    }
```
UrsKahmann: **thought:** Ideally I would put this into the `buildArticle` function so the "side effects" of adding an article (namly creating the temporal groups and now also adding the doknr entry) are always applied. But I see that touching the `buildArticle` function (adding the doknr parameter) would lead to many tests needing to add that as well 🤔  Its sad that java does not have a good concept of default parameter values...

UrsKahmann (commented):
Looks good, left a few comments, happy to discuss :)

PR #2524, opened 2026-09-15
RISDEV-11132 display expression validity on single norm page

UrsKahmann:
- show the expressions validity in the mobile & desktop TOC when on a single norm page

RISDEV-11132


andreasphil (commented):
👍

PR #2525, opened 2026-09-15
RISDEV-11132 adapt TOC typo sizes

UrsKahmann:
- add a step up in size between xs and sm. No stepup for 2xl wanted - thats why the existing typo classes are not used

RISDEV-11132


andreasphil (commented):
Not happy that we have to make an exception from the responsive typography classes here, but I'll take that to design. Code changes look good 👍

PR #2526, opened 2026-09-15
RSSDEV-12618 serve articles

ekl176:
Implements a function in the ArticleService to retrieve all versions of a specific article based on its expressionEli and eId. 
ExpressionEli and eId are used to fetch the documentNumber. The document number, being the stable identifier of an article, is used to query all versions of that article.


backend/src/main/java/de/bund/digitalservice/ris/search/service/ArticleService.java
```diff
@@ -148,6 +151,44 @@ public Optional<String> getActualEid(String expressionEli, String eidGiven) {
     return Optional.empty();
   }
 
+  /**
+   * Retrieves a List of all versions of an Article across the whole work it belongs to. A
+   * combination of expressionEli and eId is is used to retrieve the dokNr from the ris metadata.
+   * The eId is not guaranteed to be an immutable discriminator across all expressions. This dokNr
+   * is not exposed by the api yet.
```
UrsKahmann: **sueggestion:** Leave out the last sentence, if we ever add it to the api this will be forgotten and only misleading.

backend/src/main/java/de/bund/digitalservice/ris/search/service/ArticleService.java
```diff
@@ -148,6 +151,44 @@ public Optional<String> getActualEid(String expressionEli, String eidGiven) {
...
+   * Retrieves a List of all versions of an Article across the whole work it belongs to. The
+   * documentNumber is used as the article discriminator. Restricts prefix lookup to minimum-length
+   * document numbers to avoid unintended matches.
+   *
+   * @param documentNumber of a given article
+   * @return List of Article objects of the same article across all its expressions
+   */
+  private List<Article> getAllArticleVersionsByDocumentNumber(String documentNumber) {
+    if (documentNumber == null || documentNumber.length() < MIN_DOC_NUMBER_LENGTH) {
+      return List.of();
```
UrsKahmann, line 183: **question:** This hides a bit that an illegal argument was passed to this function, maybe instead throw and illeagArgument exception?

backend/src/main/java/de/bund/digitalservice/ris/search/service/ArticleService.java
```diff
@@ -148,6 +151,44 @@ public Optional<String> getActualEid(String expressionEli, String eidGiven) {
...
+    }
+
+    String documentNumberPrefix =
```
UrsKahmann: **suggestion:** Add a comment explaining that the first 21 chars of the doknr are uniquely identifying an einzelnorm across the expression. Otherwise this calculation might be hard to reason about

Or maybe even better: Instead of having the suffix lenght constant, add a constant like `UNIQUE_PREFIX_LENGHT` and then do `substring(0, UNIQUE_PREFIX_LENGHT)` which is lot more clear in my opinion and also if the doknr ever changes in length (this will probably never happen) its more fail save.

Also this way you could probably check that the docnr is at least UNIQUE_PREFIX_LENGHT long and you wouldn't need a second constant 🤔

UrsKahmann (approved):
Nice work, left a view comments, as we discussed :)

PR #2530, opened 2026-09-17
RISDEV-12620 geltungszeiten table

UrsKahmann:
- WIP


frontend/src/utils/anyDocument.ts
```diff
@@ -83,7 +83,11 @@ export function getIdentifier(document: AnyDocument): string {
 
 export function getEncodingURL(
   encoding: DocumentEncodingSchema[] | null | undefined,
-  format: string,
+  format:
```
andreasphil: nice!

frontend/src/components/documents/norms/ArticleVersionList.vue
```diff
@@ -0,0 +1,231 @@
...
+
+type VersionRow = {
+  key: string;
+  fromDate: string;
+  toDate: string;
+  contentUrl?: string;
+  disabled: boolean;
+};
+
+const columns: DataTableColumn<VersionRow>[] = [
```
andreasphil: **todo:** don't use the data table column type here even if it happens to fit what we need right now. This creates a dependency between two unrelated components that doesn't need to be there. Just add the properties you need to the `VersionRow`.

frontend/src/composables/useSingleNormVersionsHtml.ts
```diff
@@ -0,0 +1,41 @@
+import { parseDocument } from "~/utils/htmlParser.ts";
+
+export interface HtmlCacheEntry {
```
andreasphil: **todo:** remove the export here since it's only needed locally in this module. Alternatively, if you want to keep exporting it: either give it a name specific to the use case here (e.g. `SingleNormVersionHtmlCache`), or keep the name general but put the type in a different, also more general place. Similar to the data table type above: this otherwise risks creating dependencies between otherwise unrelated things.

frontend/src/composables/useSingleNormVersionsHtml.ts
```diff
@@ -0,0 +1,41 @@
...
+  html?: string;
+  error: boolean;
+}
+
+export function useSingleNormVersionsHtml() {
+  const { $risBackend } = useNuxtApp();
+  const rowsHtml = ref(new Map<string, HtmlCacheEntry>());
```
andreasphil: **todo:** this should be `reactive`, not `ref`. Refs are for references, i.e. should not be used on things that are mutated rather than replaced. `reactive` is for things that are mutated, which is what we're doing here. Even if it happens to work in this case, which one is used influences how Vue detects changes and triggers re-renders, so this can break at some point.

andreasphil (changes requested):

frontend/src/composables/useSingleNormVersionsHtml.ts
UrsKahmann: As discussed, a `ref` will make its value (in this case the map) deeply reactivity, see: [Vue docs](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#deep-reactivity). Also the docs encourage to use `ref` over `reactive` where possible. So I will leave this like it is ✌️

UrsKahmann:
Review feedback is addressed in https://github.com/digitalservicebund/ris-search/pull/2578

PR #2535, opened 2026-09-17
RISDEV-12626 Further Optimize targeted article search

andrew-cenkner-digitalservice:
* Add a dashboard that can be run to see the state after a test fails
* Add test cases
* Refactor test data creation


backend/src/integrationTest/java/de/bund/digitalservice/ris/search/service/AllDocumentsServiceIntegrationTest.java
```diff
@@ -109,4 +112,47 @@ void titlesAreNotFragmentedInTheHighlighter() {
...
+    Norm norm2 =
+        NormsTestData.buildTestNorm(
+            "StVO AusnV 3",
+            List.of("Eingangsformel", "§ 1", "§ 2"),
+            List.of("blah", "blah", "§ paragraph paragraf article artikel art abs"));
+    normsRepository.saveAll(List.of(norm1, norm2));
+    // AND that search for "§ 3 StVO blah" makes norm2 rank higher
+    var allDocs = searchAll("§ 3 StVO blah");
+    assertThat(allDocs.getFirst().getId()).isEqualTo(norm2.getId());
+
```
UrsKahmann: **question:** Why is norm2 ranked higher even though it doesn't even have a `§ 3` and the abbreviation also doesn't match? If this is correct I think a short explanation why would be helpful as a comment, because otherwise this is not directly obvious 🤔

backend/src/integrationTest/java/de/bund/digitalservice/ris/search/service/NormsServiceIntegrationTest.java
```diff
@@ -188,4 +204,43 @@ void shouldFilterOnlyOnFullKeywordMatchesByRisAbbreviation() {
...
+            List.of("§ 1", "§ 2", "§ 3"),
+            List.of("blah", "blah", "§ paragraph paragraf article artikel art abs"));
+    Norm norm2 =
+        NormsTestData.buildTestNorm(
+            "StVO AusnV 3",
+            List.of("Eingangsformel", "§ 1", "§ 2"),
+            List.of("blah", "blah", "§ paragraph paragraf article artikel art abs"));
+    normsRepository.saveAll(List.of(norm1, norm2));
+    // AND that search for "§ 3 StVO blah" makes norm2 rank higher (because a norm2 has '3' in the
+    // abbreviation)
```
UrsKahmann: **praise:** Ahh here you have the explanation, great! Would be nice if you add it to the other test as well 🙏

UrsKahmann (approved):
Looks good, only a tiny nitpick regarding the one comment in a test ;)
I like where the test structure and setup is going 🚀 

Regarding the new docker compose file: I guess ideally we have as little compose files as possible, but for this I also don't see a better way without. While at it i was wondering if opensearch compose files should maybe moved to the backend as they are more backend related and then the projects root directory is less cluttered 🤔 But just a though :)

PR #2537, opened 2026-09-18
RISDEV-0000 responsive typo cleanup

andreasphil:
This PR simplifies the many exceptions we made from our responsive typography system:

- adds a new `link2` util, mirroring the existing `label2`, and updates all relevant locations
- renames the existing `link` to `link1`
- creates lead heading styles for the new landing page as an exception in the typography CSS
- adds a new `label1-compact` variant for all instances where we need a `label1` except on the biggest breakpoint
- replaces instances where we manually repeated the responsive typo definition 1:1
- adds missing stories
- fixes missing border for Gesamtausgabe links in Chrome


frontend/src/components/documents/TableOfContents.vue
```diff
@@ -60,7 +60,7 @@ const drawerId = useId();
         </div>
         <span
           v-if="subheading && subheadingAddition"
-          class="ris-label2-regular line-clamp-1"
+          class="typo-label2-regular line-clamp-1"
```
UrsKahmann, line 63: **question:** With this change, the typo is `ris-label1-regular` for `2xl` screens which it wasn't before.. I think it looks good, but should we maybe check with design? Or do you think its okay to just do it :)

UrsKahmann (commented):
Thanks for making this more consistent 🙌

frontend/src/components/documents/TableOfContents.vue
andreasphil: correct, but that will never be visible to users since we don't show this part on large screens. On smaller screens it will be identical and I'd prefer using the `typo-` classes where possible.

UrsKahmann: Ahh true 🤦  I didn't say anything :D

PR #2538, opened 2026-09-18
RISDEV-12577 Improve tests

andrew-cenkner-digitalservice:
* Move some beans to ContainersIntegrationBase to lower the number of spring contexts we have. A bean under test can always be a @MockitoSpyBean. It just allows the user to use `verify` on the bean, but otherwise the behavior isn't changed.
* Minor cleanup in ContainersIntegrationBase


backend/src/integrationTest/java/de/bund/digitalservice/ris/search/config/ContainersIntegrationBase.java
```diff
@@ -92,6 +91,20 @@ public class ContainersIntegrationBase {
...
+  @Autowired protected CaseLawService caseLawService;
+  @Autowired protected NormsService normsService;
+  @Autowired protected LiteratureService literatureService;
+  @Autowired protected AdministrativeDirectiveService administrativeDirectiveService;
+
+  @Qualifier("normsBulkExportService")
+  @MockitoSpyBean
+  protected BulkExportService normsBulkExportService;
+
+  @MockitoSpyBean protected PostHogService postHogService;
```
UrsKahmann, line 105: **question:** This is only used in one test class right? I don't really see the value of moving it to the base class then.. This just adds another layer of indirection to the `FeedbackControllerIntegrationTest`.

UrsKahmann (approved):
Looks good 🚀 I only left one question :)

backend/src/integrationTest/java/de/bund/digitalservice/ris/search/config/ContainersIntegrationBase.java
andrew-cenkner-digitalservice: It's a trade off, but a bit hard to explain in a comment. The reason is by putting it in a parent we have one less spring context in our test suite. Let's discuss more in the next lean coffee.

PR #2539, opened 2026-09-18
Risdev 12618 article version manifestations grouped

ekl176:


backend/src/integrationTest/java/de/bund/digitalservice/ris/search/repository/ArticlesRepositoryCustomImplIntegrationTest.java
UrsKahmann, line 1: **praise:** Really nice tests, well structured and the comments really help to understand what is tested. One minor improvement: Use simplified doknummern to make it easier to see the differences. E.g. DKNR000000000000000010001, DKNR000000000000000020001  etc.

backend/src/main/java/de/bund/digitalservice/ris/search/mapper/NormLdmlToOpenSearchMapper.java
UrsKahmann, line 1: **praise:** Really nice simplification with the normContext 🚀

backend/src/main/java/de/bund/digitalservice/ris/search/models/opensearch/ArticleWithExpressions.java
UrsKahmann, line 1: **suggestion:** For someone not knowing the concept with the document number, this comment is really confusing. I would instead refer to "article version". E.g.:
```java
/**
 * Carries an {@link Article} representing one distinct version together with the
 * expressionElis of every expression that version of the article occurs in.
 *
 * @param article the representative for this version of article
 * @param expressionElis expressionElis of every expression the article occurs in
 */
```

backend/src/main/java/de/bund/digitalservice/ris/search/repository/opensearch/ArticlesRepositoryCustom.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Custom {@link ArticlesRepository} operations that can't be expressed as derived Spring Data query
+ * methods.
+ */
+public interface ArticlesRepositoryCustom {
+
+  /**
+   * Retrieves articles whose document number starts with the given prefix, collapsed on the full
+   * document number so that only one {@link Article} per distinct document number is returned. Each
+   * result also carries the expressionElis of every expression that document number occurs in.
```
UrsKahmann: **suggestion:** Rephrase to "... of every expression that article occurs in."

backend/src/main/java/de/bund/digitalservice/ris/search/repository/opensearch/ArticlesRepositoryCustomImpl.java
```diff
@@ -0,0 +1,111 @@
...
+
+/** Repository with custom Article operations */
+public class ArticlesRepositoryCustomImpl implements ArticlesRepositoryCustom {
+
+  private static final String EXPRESSIONS_INNER_HIT_NAME = "expressions";
+
+  private static final String DISTINCT_DOCUMENT_NUMBERS_AGGREGATION_NAME =
+      "distinct_document_numbers";
+
+  private static final int MAX_EXPRESSIONS_PER_ARTICLE = 100;
```
UrsKahmann, line 32: **question:** Does this mean we only get a max of 100 expressions per article? E.g. if an article doesn't change at all and the norm has more than 100 expressions -> some are not returned?

UrsKahmann (approved):
Really really nice work 🚀

backend/src/main/java/de/bund/digitalservice/ris/search/repository/opensearch/ArticlesRepositoryCustomImpl.java
ekl176: When not setting any maximum the default innerHit size will be 3. The maximum allowed by default is 100, so we can't increase this further without changing index configuration. 
We might even want to decrease this in the future depending if a smaller result set proves to be sufficient

